PR #4, opened 2022-08-03
Add upstream codebase

qbarrand:
This PR:
- merges code from upstream;
- changes the import path;
- changes all references from the old repo and image registry to the new one;
- vendors dependencies.


yevgeny-shnaidman:
couple of comments:
1)I see that Dockerfile.kmm-operator-build and Dockerfile.merge still use openshift images. Are we going to fix it now, or just open a ticket a fixed afterwards? ( i am good with both ways, as long as we don't forget it)
2) vendor directory - do we need it in the upstream?

qbarrand:
@yevgeny-shnaidman 
>  1)I see that Dockerfile.kmm-operator-build and Dockerfile.merge still use openshift images. Are we going to fix it now, or just open a ticket a fixed afterwards? ( i am good with both ways, as long as we don't forget it)

This is midstream, so I think it's fine to have Red Hat images here.

>  2) vendor directory - do we need it in the upstream?

This is midstream, so I think we need it unfortunately.

yevgeny-shnaidman (approved):

PR #6, opened 2022-08-05
Make .gitignore patterns more specific

qbarrand:
`.gitignore` contained `manager`, which made git ignore any directory named `manager` in the repository, including a `controller-runtime` package.


ybettan:
Can we please update the commit message to contain the explanation you put in the PR description? (the commit message itself don't tell too much about the change).

Besides that, lgtm.

PR #7, opened 2022-08-05
Download golangci-lint instead of installing it in a vendored environment

qbarrand:
Among the consequences of vendoring dependencies is the fact that we cannot `go install` any package from the internet.
We have to either:
- list the code of the package as a dependency and vendor it, even though the operator does not use it, or;
- install the binary we need through other means.

This PR installs `golangci-lint` by downloading it from the GitHub release.


yevgeny-shnaidman:
why have you changed it to a PATH location instead of under the project?

qbarrand:
@yevgeny-shnaidman not sure I understand your comment. The binary still goes into the same `bin` directory.

ybettan (approved):

PR #8, opened 2022-08-08
Do not install envtest when running unit tests

qbarrand:
`operator-sdk` generates a `Makefile` that downloads [EnvTest](https://book.kubebuilder.io/reference/envtest.html) by default for unit tests, although we do not use it.
Installing unvendored dependencis is a problem in CI, so remove the dependency on EnvTest altogether.


ybettan:
Commit message is empty.

ybettan (approved):

ybettan (changes requested):
Can you please fill the commit message according to the contribution guide?

PR #12, opened 2022-08-09
Draft: Install golangci-lint in the lint CI script

qbarrand:


qbarrand:
We will solve this problem in https://github.com/openshift/release/pull/31190 instead.

PR #14, opened 2022-08-11
Add required variables for the lint CI job

qbarrand:
These variables are necessary for `golangci-lint` to run correctly.

/cc @ybettan


ybettan:
Missing a commit message. Besides that lgtm.

PR #17, opened 2022-08-16
Import the new CRD from upstream

qbarrand:
Downstream cherry-pick of https://github.com/qbarrand/oot-operator/pull/127.


api/v1beta1/groupversion_info.go
```diff
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1alpha1 contains API Schema definitions for the kmm v1alpha1 API group
+// Package v1beta1 contains API Schema definitions for the kmm v1beta1 API group
 //+kubebuilder:object:generate=true
 //+groupName=kmm.sigs.k8s.io
```
ybettan, line 19: Are we keeping the `sigs` name m/s as well?

api/v1beta1/groupversion_info.go
```diff
@@ -26,7 +26,7 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "kmm.sigs.k8s.io", Version: "v1alpha1"}
+	GroupVersion = schema.GroupVersion{Group: "kmm.sigs.k8s.io", Version: "v1beta1"}
```
ybettan, line 29: Shouldn't we use another name for the m/s repo? I don't mind if we don't, just asking.

ybettan:
I see some references to `sigs` in the PR (resource group name and file names). Is that intended?

api/v1beta1/groupversion_info.go
qbarrand: I don't think so. I think it would be convenient for people to use the same CR for both vanilla Kubernetes and OCP. WDYT?

ybettan: I am for all what is simpler. Let merge that way.

PR #23, opened 2022-08-23
Add BuildConfig support

qbarrand:
Remove the Job-based build manager.
Use the pull secrets from the `builder` ServiceAccount by default.

/cc @ybettan @yevgeny-shnaidman @enriquebelarte @fabiendupont


internal/auth/auth.go
```diff
@@ -8,11 +8,31 @@ import (
...
+}
+
+type registryAuthGetterFactory struct{}
+
+func NewRegistryAuthGetterFactory() RegistryAuthGetterFactory {
+	return registryAuthGetterFactory{}
+}
+
+func (registryAuthGetterFactory) NewRegistryAuthGetter(client client.Client, namespacedName types.NamespacedName) RegistryAuthGetter {
+	return NewRegistryAuthGetter(client, namespacedName)
```
yevgeny-shnaidman, line 29: Why do we need NewRegistryAuthGetter and NewServiceAccountAuthGetter functions? why not just use their implementation in the member functions? I think that after this commit all the code will simply start using RegisterAuthGetterFactory, no?

internal/build/buildconfig/maker.go
```diff
@@ -0,0 +1,124 @@
...
+					Type:       buildv1.BuildSourceDockerfile,
+				},
+				Strategy: buildv1.BuildStrategy{
+					Type: buildv1.DockerBuildStrategyType,
+					DockerStrategy: &buildv1.DockerBuildStrategy{
+						BuildArgs: EnvVarsFromKMMBuildArgs(buildArgs),
+						Volumes:   BuildVolumesFromBuildSecrets(buildConfig.Secrets),
+					},
+				},
+				Output: buildv1.BuildOutput{
```
yevgeny-shnaidman, line 64: Are we going to use ImageStream? Is it next stage?

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+//go:generate mockgen -source=manager.go -package=buildconfig -destination=mock_manager.go
+
+type OpenShiftBuildsHelper interface {
+	GetBuildConfig(ctx context.Context, mod kmmv1beta1.Module, targetKernel string) (*buildv1.BuildConfig, error)
+	GetLatestBuild(ctx context.Context, namespace, buildConfigName string) (*buildv1.Build, error)
+}
+
+type openShiftBuildsHelper struct {
```
yevgeny-shnaidman, line 133: Why do we need this interface?

internal/build/buildconfig/maker.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func NewMaker(helper build.Helper, scheme *runtime.Scheme) Maker {
+	return &maker{
+		helper: helper,
+		scheme: scheme,
+	}
+}
+
+func (m *maker) MakeBuildConfig(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string) (*buildv1.BuildConfig, error) {
```
ybettan, line 34: I think that `buildConfig` is a confusing name, at first sight, it make me think this is an OCP buildConfig object but indeed it is a KMM build object.

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+	coreClient      kubernetes.Interface
+	helper          build.Helper
+	maker           Maker
+	ocpBuildsHelper OpenShiftBuildsHelper
+	registry        registry.Registry
+}
+
+func NewManager(
+	client client.Client,
+	coreClient kubernetes.Interface,
```
ybettan, line 38: What is the difference between `client.Client` nad `kubeneretens.Interface`?

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		registryAuthGetter = bcm.authFactory.NewRegistryAuthGetter(bcm.client, namespacedName)
+	} else {
+		logger.Info("Getting pull secret from Service Account", "service account", builderServiceAccountName)
+
+		registryAuthGetter = bcm.authFactory.NewServiceAccountRegistryAuthGetter(bcm.coreClient, mod.Namespace, builderServiceAccountName)
+	}
+
+	imageAvailable, err := bcm.registry.ImageExists(ctx, m.ContainerImage, config.Pull, registryAuthGetter)
```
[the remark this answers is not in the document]
ybettan: Do we want to create a k8s object (BuildConfig) even if there is no need to? As a user, by seeing a BuildConfig in the cluster I would assume that a build was required...

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not check if the image is available: %v", err)
+	}
+
+	if imageAvailable {
+		return build.Result{Status: build.StatusCompleted, Requeue: false}, nil
```
ybettan, line 84: WDYT about adding a new status `StatusSkipped` to `build.Result`? Using `StatusCompleted` doesn't feels right here.

internal/build/buildconfig/manager.go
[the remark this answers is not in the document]
ybettan: Line 109 is assuming that a `BuildConfig` and some `Build` objects where already created which may not always be the case.

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+		if err = bcm.client.Create(ctx, buildConfig); err != nil {
+			return build.Result{}, fmt.Errorf("could not create Job: %v", err)
+		}
+
+		return build.Result{Status: build.StatusCreated, Requeue: true}, nil
+	}
+
+	b, err := bcm.ocpBuildsHelper.GetLatestBuild(ctx, mod.Namespace, buildConfig.Name)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not find the latest build: %v", err)
```
ybettan, line 111:
```suggestion
		return build.Result{}, fmt.Errorf("could not find the latest build of buildconfig/%s in namespace %s: %v", buildConfig.Name, mode.Namespace, err)
```

internal/build/buildconfig/manager.go
ybettan: Same here.

Last time we talked about it the conclusions were:
* We shouldn't create new interfaces unless it functionality is being used in another package
* If it functionality is big enough, and we want to test it separately and mock it, it should be extracted to another package, tested separately, and imported by a consumer.

internal/build/buildconfig/maker.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if err := controllerutil.SetControllerReference(&mod, &bc, m.scheme); err != nil {
+		return nil, fmt.Errorf("could not set the owner reference: %v", err)
+	}
+
+	return &bc, nil
+}
+
+func EnvVarsFromKMMBuildArgs(args []kmmv1beta1.BuildArg) []v1.EnvVar {
```
ybettan, line 84:
```suggestion
func envVarsFromKMMBuildArgs(args []kmmv1beta1.BuildArg) []v1.EnvVar {
```

internal/build/buildconfig/maker.go
```diff
@@ -0,0 +1,124 @@
...
+	ev := make([]v1.EnvVar, 0, len(args))
+
+	for _, ba := range args {
+		ev = append(ev, v1.EnvVar{Name: ba.Name, Value: ba.Value})
+	}
+
+	return ev
+}
+
+func BuildVolumesFromBuildSecrets(secrets []v1.LocalObjectReference) []buildv1.BuildVolume {
```
ybettan, line 98:
```suggestion
func buildVolumesFromBuildSecrets(secrets []v1.LocalObjectReference) []buildv1.BuildVolume {
```

internal/build/buildconfig/manager.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, errNoMatchingBuild
+	}
+
+	latest := buildv1.Build{}
+
+	for i := 0; i < len(builds.Items); i++ {
+		if b := builds.Items[i]; b.CreationTimestamp.After(latest.CreationTimestamp.Time) {
+			latest = b
+		}
+	}
```
ybettan, line 184:
```suggestion
	for _, b := range(builds.Items) {
		if b.CreationTimestamp.After(latest.CreationTimestamp.Time) {
			latest = b
		}
	}
```

internal/build/buildconfig/maker.go
qbarrand: The user only specifies a container image name as destination; it is not easy for us to determine if that image name is from an ImageStream or not.
The usual authentication against the ImageStream works even though you are using an image name, so I don't think there is any downside right now.

internal/build/buildconfig/manager.go
qbarrand: @ybettan `kubernetes.Interface` is from `client-go` while `client.Client` is a generic overlay on top from controller-runtime.
I would not have our own client like we had in SRO, which would bring a lot of code. I would stick to `client.Client` as much as possible.

internal/build/buildconfig/manager.go
qbarrand: Why doesn't it feel right? If we add `StatusSkipped`, I believe we'll move straight from `build.StatusInProgress` to `build.StatusSkipped`.

internal/build/buildconfig/manager.go
qbarrand: Good catch. This works because `latest` is not a pointer.

internal/build/buildconfig/manager.go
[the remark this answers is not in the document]
yevgeny-shnaidman: in case we remove the image verification, and just check the BuildConfig status, then we don't need registryAuthGetter

internal/auth/auth.go
yevgeny-shnaidman: i understand the need for Factory, i mean that Factory New... functions are calling additional functions ( not member functions of Factory). Why do we need this 2 layer call?

internal/build/buildconfig/maker.go
yevgeny-shnaidman: So customer will be responsible for creating image stream, right?

internal/build/buildconfig/manager.go
ybettan: > Private interface

Not sure what you mean by "Private interface". But I would create a new package (maybe in the `buildconfig` package - something like `buildconfig.helper` package).

That package will hold the `OpenshiftBuildsHelper` interface and you consume it in `buildconfig` - all the rest, including tests, should remain the same.

yevgeny-shnaidman: i am not sure if we need to multiply the packages, but maybe move this interface to a different file under the same package?

internal/build/buildconfig/maker.go
ybettan: Sure.

internal/build/buildconfig/maker.go
qbarrand: Yes.

internal/build/buildconfig/manager.go
ybettan: I may be missing something here but isn't that case cover the story in which the user has pre-build his driver? If that the case, how will we reach `build.STatusInProgress`?

qbarrand: If the image is pre-built, then there won't be any `build` section for that mapping in the CRD and the reconciler will not call this method.

internal/build/buildconfig/manager.go
ybettan: But then there we still export an interface which is only used internally in the package.

internal/build/buildconfig/manager.go
ybettan: So why do we check if the image is available to begin with? In which cases the image will already exist?

internal/build/buildconfig/manager.go
ybettan: A non exported interface (maybe this is what @qbarrand meant by "private interface") is also an option but then how will you test it? Are you OK with testing private interfaces? Also I am not sure we will be able to mock private interfaces.

yevgeny-shnaidman: i am not sure if there is such thing as private interfaces... What does it mean? That it is not public?  I think we will be able to test it if the we define unit test in the same package, as we usually do.

internal/build/buildconfig/manager.go
yevgeny-shnaidman: @qbarrand i am not sure that we can count on build section not present in case image is prebuilt. Build section is a fallback. Maybe we need to move the image existence to an upper layer, prior to entering the build flow.  Mayb e it should be defined in the HandleBuild function in controller, or even before calling handleBuild

internal/build/buildconfig/manager.go
ybettan: I don't mind your approach either, I tried to find a win-win solution.
In that case it might be easy to test because those private functions have very lean logic and underlying API k8s calls that can be mock so it that specific case, private function should be OK as well.

In other cases, when the logic isn't very lean, I think we should extract to another interface and export, otherwise, tests are a nightmare.

internal/build/buildconfig/maker.go
qbarrand: Ended up using `kmmBuild` as `build` is a package name already.

internal/build/manager.go
```diff
@@ -22,5 +22,6 @@ type Result struct {
 //go:generate mockgen -source=manager.go -package=build -destination=mock_manager.go
 
 type Manager interface {
+	IsOutputPresent(ctx context.Context, mod kmmv1beta1.Module, containerImage string) (bool, error)
```
yevgeny-shnaidman, line 25: Why do we need this separate API in the Build Interface. Why not use directly the registry interface? Logically checking image existence is not related to build, we can use this API  in other flows also (not related to build)

internal/build/buildconfig/maker.go
qbarrand: Done.

internal/build/buildconfig/maker.go
qbarrand: Done.

PR #31, opened 2022-08-28
Revert "Bump k8s.io/kubectl from 0.24.4 to 0.25.0"

ybettan:
Reverts rh-ecosystem-edge/kernel-module-management#28

OCP 4.12 is using k8s 1.24 so keep 0.24.4 seems as a better fit here.


ybettan:
As @qbarrand has pointed out, this bump is going further in versions than what OCP 4.12 delivers (at least in nightlies) so it seems better to keep the 0.24.4 versions.

PR #39, opened 2022-09-08
Separate PullOption for build process and image verification

qbarrand:
Currently we have pull option in the Build section of the Module API. It is used both
for configuring the pull option passed to the build configuration (kaniko)
and also to the function that verifies images existence (pulling its manifest).
We should separate them into 2 different entities:
1) present in ModuleLoader spec - used for image contents inspection and existence verification
2) present in the Build - used for specifying pull options of the base images of the build process


qbarrand:
Upstream commit: https://github.com/kubernetes-sigs/kernel-module-management/commit/1b7f0b3e3594f662749af6e2b76b3ffde37aaf16

PR #46, opened 2022-09-19
Add permissions to module loader pod (#151)

yevgeny-shnaidman:
To be able to load a module, more privileges are required. This change sets the SELinux type of the pod to `spc_t` and the user to `root`. It also explicitely disable privilege escalation.

This has been tested with a pod ServiceAccount that can use the `privileged` SCC.

Reference: [Introducing a Super Privileged Container Concept](https://developers.redhat.com/blog/2014/11/06/introducing-a-super-privileged-container-concept)

Signed-off-by: Fabien Dupont <fdupont@redhat.com>


yevgeny-shnaidman:
the original PR:
https://github.com/qbarrand/oot-operator/pull/151

PR #53, opened 2022-09-26
Automate module loader and device plugin SAs bindings to privileged SCC

mresvanis:
This PR adds the changes required to automate the binding of the default module loader and device plugin SAs to the `privileged` [SecurityContextConstraints](https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html), in order for users to seamlessly have their `Module`'s pods [admitted](https://connect.redhat.com/en/blog/important-openshift-changes-pod-security-standards) on OCP 4.12+ (due to the enforced K8S [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/)).

> This takes place only when the `Module` CR does not define the respective `ServiceAccountName` fields of the `Module.Spec.ModuleLoader` and/or the `Module.Spec.DevicePlugin` specs.

Specifically, it adds:
- a module loader `ClusterRole`, which allows the use of the `privileged` SCC, to the OLM manifests
- a device plugin `ClusterRole`, which allows the use of the `privileged` SCC, to the OLM manifests
- controller permissions to use the privileged SCC
- controller permissions to bind the afore-mentioned `ClusterRole`s with the privileged SCC to SAs
- a `RoleBinding` of the module loader `ClusterRole` to the module loader SA
- a `RoleBinding` of the device plugin `ClusterRole` to the device plugin SA

[MGMT-11574](https://issues.redhat.com//browse/MGMT-11574)
  
Signed-off-by: Michail Resvanis <mresvani@redhat.com>

### DIFF
--- a/config/rbac/module_loader_cluster_role.yaml
+++ b/config/rbac/module_loader_cluster_role.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: module-loader
+rules:
+- apiGroups:
+  - security.openshift.io
+  resourceNames:
+  - privileged
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -94,6 +94,9 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=serviceaccounts,verbs=create;delete;get;list;patch;watch
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=use,resourceNames=privileged
+//+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=create;delete;get;list;patch;watch
+//+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterroles,verbs=bind,resourceNames=module-loader;device-plugin
 //+kubebuilder:rbac:groups="build.openshift.io",resources=buildconfigs,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="build.openshift.io",resources=builds,verbs=get;list;watch
 
@@ -118,13 +121,13 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	r.setKMMOMetrics(ctx)
 
 	if mod.Spec.ModuleLoader.ServiceAccountName == "" {
-		if err := r.rbacAPI.CreateModuleLoaderServiceAccount(ctx, *mod); err != nil {
-			return res, fmt.Errorf("could not create module-loader's ServiceAccount: %w", err)
+		if err := r.rbacAPI.CreateModuleLoaderRBAC(ctx, *mod); err != nil {
+			return res, fmt.Errorf("could not create module-loader's RBAC: %w", err)
 		}
 	}
 	if mod.Spec.DevicePlugin != nil && mod.Spec.DevicePlugin.ServiceAccountName == "" {
-		if err := r.rbacAPI.CreateDevicePluginServiceAccount(ctx, *mod); err != nil {
-			return res, fmt.Errorf("could not create device-plugin's ServiceAccount: %w", err)
+		if err := r.rbacAPI.CreateDevicePluginRBAC(ctx, *mod); err != nil {
+			return res, fmt.Errorf("could not create device-plugin's RBAC: %w", err)
 		}
 	}
 
--- a/internal/rbac/rbac.go
+++ b/internal/rbac/rbac.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -17,8 +18,8 @@ import (
 //go:generate mockgen -source=rbac.go -package=rbac -destination=mock_rbac.go
 
 type RBACCreator interface {
-	CreateModuleLoaderServiceAccount(ctx context.Context, mod kmmv1beta1.Module) error
-	CreateDevicePluginServiceAccount(ctx context.Context, mod kmmv1beta1.Module) error
+	CreateModuleLoaderRBAC(ctx context.Context, mod kmmv1beta1.Module) error
+	CreateDevicePluginRBAC(ctx context.Context, mod kmmv1beta1.Module) error
 }
 
 type rbacCreator struct {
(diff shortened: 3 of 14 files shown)


config/rbac/module_loader_cluster_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: module-loader
```
qbarrand, line 4:
```suggestion
  name: kmm-operator-module-loader
```

config/rbac/module_loader_scc.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: module-loader
```
qbarrand, line 4:
```suggestion
  name: kmm-operator-module-loader
```

controllers/module_reconciler.go
```diff
@@ -118,6 +124,18 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	r.setKMMOMetrics(ctx)
 
+	if mod.Spec.ModuleLoader.ServiceAccountName == "" {
+		logger.Info("Handle default driver container ServiceAccount")
+		if err := r.handleDriverContainerServiceAccount(ctx, mod); err != nil {
+			return res, fmt.Errorf("could not handle default driver container ServiceAccount: %w", err)
```
qbarrand, line 130:
```suggestion
		logger.Info("Handle default module loader ServiceAccount")
		if err := r.handleModuleLoaderServiceAccount(ctx, mod); err != nil {
			return res, fmt.Errorf("could not handle default module loader ServiceAccount: %w", err)
```

controllers/module_reconciler.go
```diff
@@ -118,6 +124,18 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		}
+
+		logger.Info("Handle default driver container RoleBinding")
+		if err := r.handleDriverContainerRoleBinding(ctx, mod); err != nil {
+			return res, fmt.Errorf("could not handle default driver container RoleBinding: %w", err)
```
qbarrand, line 135:
```suggestion
		logger.Info("Handle default module loader RoleBinding")
		if err := r.handleModuleLoaderRoleBinding(ctx, mod); err != nil {
			return res, fmt.Errorf("could not handle default module loader RoleBinding: %w", err)
```

controllers/module_reconciler.go
```diff
@@ -298,6 +316,48 @@ func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1
 	return imageAvailable, nil
 }
 
+func (r *ModuleReconciler) handleDriverContainerServiceAccount(ctx context.Context, mod *kmmv1beta1.Module) error {
+	logger := log.FromContext(ctx)
+
+	sa := &v1.ServiceAccount{}
+	fmt.Println("Before rbacAPI.GetDriverContainerServiceAccount")
```
qbarrand, line 323: Leftover?

controllers/module_reconciler.go
```diff
@@ -298,6 +316,48 @@ func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1
...
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, r.Client, sa, func() error {
+		return r.rbacAPI.SetDriverContainerServiceAccountAsDesired(ctx, sa, *mod)
+	})
+
+	if err == nil {
+		logger.Info("Reconciled driver container ServiceAccount", "name", sa.Name, "result", opRes)
+	}
+
+	return err
```
qbarrand, line 338: Nit: could we use a more traditional flow here?
```suggestion
	if err != nil {
		return fmt.Errorf("could not reconcile ServiceAccount: %v", err)
	}
	
	logger.Info("Reconciled driver container ServiceAccount", "name", sa.Name, "result", opRes)
	
	return nil
```

controllers/module_reconciler.go
```diff
@@ -298,6 +316,48 @@ func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1
 	return imageAvailable, nil
 }
 
+func (r *ModuleReconciler) handleDriverContainerServiceAccount(ctx context.Context, mod *kmmv1beta1.Module) error {
+	logger := log.FromContext(ctx)
+
+	sa := &v1.ServiceAccount{}
+	fmt.Println("Before rbacAPI.GetDriverContainerServiceAccount")
+	err := r.rbacAPI.GetDriverContainerServiceAccount(ctx, *mod, sa)
```
qbarrand, line 324: Why do we need to get the ServiceAccount first?

controllers/module_reconciler.go
```diff
@@ -298,6 +316,48 @@ func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1
...
+	})
+
+	if err == nil {
+		logger.Info("Reconciled driver container ServiceAccount", "name", sa.Name, "result", opRes)
+	}
+
+	return err
+}
+
+func (r *ModuleReconciler) handleDriverContainerRoleBinding(ctx context.Context, mod *kmmv1beta1.Module) error {
```
qbarrand, line 341: Should we group the `RoleBinding` and `ServiceAccount` management into just one function?

internal/rbac/rbac.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+//go:generate mockgen -source=rbac.go -package=rbac -destination=mock_rbac.go
+
+type RBACCreator interface {
+	GetDriverContainerServiceAccount(ctx context.Context, mod kmmv1beta1.Module, sa *corev1.ServiceAccount) error
+	GetDriverContainerRoleBinding(ctx context.Context, mod kmmv1beta1.Module, rb *rbacv1.RoleBinding) error
+	SetDriverContainerServiceAccountAsDesired(ctx context.Context, sa *corev1.ServiceAccount, mod kmmv1beta1.Module) error
+	SetDriverContainerRoleBindingAsDesired(ctx context.Context, rb *rbacv1.RoleBinding, mod kmmv1beta1.Module) error
+}
```
qbarrand, line 25: Wondering if this interface should be simplified to just:
- `EnsureModuleLoaderRBAC`
- `EnsureDevicePluginRBAC`

This would eliminate the need for `handle*` private methods and reduce the number of calls to mock.

internal/daemonset/daemonset.go
```diff
@@ -204,6 +204,11 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 		container.VolumeMounts = append(container.VolumeMounts, firmwareVolumeMount)
 	}
 
+	serviceAccountName := mod.Spec.ModuleLoader.ServiceAccountName
```
qbarrand, line 207: Wondering if the SA management also belongs upstream - without the SCC. WDYT?

qbarrand (commented):
Thank you!

config/rbac/module_loader_cluster_role.yaml
mresvanis: You revealed pretty fast a pain point regarding the naming and the `kustomize` manifests, with which I would like some help :)

#### Context
We have setup the [namePrefix](https://github.com/mresvanis/rh-kernel-module-management/blob/main/config/default/kustomization.yaml#L9), so all known to kustomize resources are named with the `kmm-operator-` prefix by default. But at the same time, I didn't find a way to use this transformer for the `resourceNames` field of the `ClusterRole`.  So I had to manually put the `kmm-operator-` prefix to the `resourceNames` field of the `ClusterRole`.

#### Ideas?
As this might break the whole thing if someone chooses to replace the `namePrefix` when running `kustomize`, do you think we could maybe opt-out these manifests from the `namePrefix` transformer?

controllers/module_reconciler.go
mresvanis: :see_no_evil:

controllers/module_reconciler.go
mresvanis: Absolutely :+1:

controllers/module_reconciler.go
mresvanis: I was leaning towards this direction myself actually, I'll give it another go. It's been smelling since I pressed the first keystroke :)

internal/daemonset/daemonset.go
mresvanis: This is a very good point, I do think that we should bring the SA management upstream as well. 

> It is a good practice to have a SA  per workload and we shouldn't rely on the default SA (when the KMM user don't define any SA). The (admin) users can change the default SA to fulfill other needs for different types of workloads at any point.

controllers/module_reconciler.go
mresvanis: Very good point, too hasty of a decision there! I will change it.

internal/rbac/rbac.go
mresvanis: Agreed! It smelled funny to me as well :+1:

config/rbac/module_loader_cluster_role.yaml
mresvanis: Thanks for the help! The [solution](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/53/files#diff-006ba62583b26e070f2015bafbd541d453d58f2d3136ef65522d835f8b002ce8) you pointed out works as expected.

config/rbac/module_loader_scc.yaml
mresvanis: This is also resolved via https://github.com/rh-ecosystem-edge/kernel-module-management/pull/53#discussion_r981120263.

controllers/module_reconciler.go
mresvanis: Refactored the internal RBAC API and changed all `driver container` references to `module loader` ones.

controllers/module_reconciler.go
mresvanis: Same as in https://github.com/rh-ecosystem-edge/kernel-module-management/pull/53#discussion_r981121561.

internal/rbac/rbac.go
```diff
@@ -33,52 +34,142 @@ func NewCreator(client client.Client, scheme *runtime.Scheme) RBACCreator {
...
+
+func (rc *rbacCreator) CreateModuleLoaderRBAC(ctx context.Context, mod kmmv1beta1.Module) error {
 	logger := log.FromContext(ctx)
 
+	opRes, err := rc.createModuleLoaderServiceAccount(ctx, mod)
+	if err != nil {
+		return fmt.Errorf("cound not create module-loader's ServiceAccount: %w", err)
+	}
+	logger.Info("Created module-loader's ServiceAccount", "name", GenerateModuleLoaderServiceAccountName(mod), "result", opRes)
```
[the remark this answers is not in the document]
mresvanis: I also thought that piece smelled, so maybe 8b2de45a9c697a23e1505f9bc213b06eb092e6b6 could be a bit more practical?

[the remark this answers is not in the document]
qbarrand: I think it looks much cleaner. Is there still a need for `GenerateModuleLoaderServiceAccountName` and `GenerateDevicePluginServiceAccountName` then?

yevgeny-shnaidman:
a general question: do we want this code in the downstream?
if we are allowing customer to specify the SA in the Module CR, then maybe it is better to let him define all the necessary connection? @qbarrand , @mresvanis

qbarrand:
@yevgeny-shnaidman downstream is actually the motivation for this change.
We want module loader DaemonSets to be able to run out of the box - that means binding their SA to the right SCC.
Right now, we have to run `oc adm policy add-scc-to-user privileged -z default` for the DaemonSet to run - this change automates that.

yevgeny-shnaidman:
> @yevgeny-shnaidman downstream is actually the motivation for this change. We want module loader DaemonSets to be able to run out of the box - that means binding their SA to the right SCC. Right now, we have to run `oc adm policy add-scc-to-user privileged -z default` for the DaemonSet to run - this change automates that.

@qbarrand when i talked to @mresvanis he said that Openshift prefers that customer/user will make those changes ( just as we let the customer add labels to the namespace for pod security). In addition, if we do all this heavy lifting, why do we need the SA in the module CR?

mresvanis:
> > @yevgeny-shnaidman downstream is actually the motivation for this change. We want module loader DaemonSets to be able to run out of the box - that means binding their SA to the right SCC. Right now, we have to run `oc adm policy add-scc-to-user privileged -z default` for the DaemonSet to run - this change automates that.
> 
> @qbarrand when i talked to @mresvanis he said that Openshift prefers that customer/user will make those changes ( just as we let the customer add labels to the namespace for pod security). In addition, if we do all this heavy lifting, why do we need the SA in the module CR?

Hey folks, let me also try to clarify, from my point of view, the motivation and trade-offs considered for this PR. 

OpenShift 4.12 (Kubernetes 1.25) introduced [the enforcement of Pod Security Standards](https://connect.redhat.com/en/blog/important-openshift-changes-pod-security-standards) (via namespace labels), which lead to workloads having to adhere to 1 of the 3 standards. In addition to the PSS, workloads in OpenShift also need to adhere to a `SecurityContextConstraints`. The OpenShift [label synchronization component](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/pod-security-admission-autolabeling.md#pod-security-admission-label-synchronization) can add the respective PSS namespace labels, depending on the SCCs used in a namespace. The latter makes it a lot easier for OpenShift users to not have to deal at all with PSS, but instead keep using SCCs.

The KMM workloads (i.e. module-loader, device-plugin) belong to the `Privileged` PSS and as such, downstream KMM users need to do one of the following to successfully have their KMM DS pod admitted:

- create ServiceAccounts, RoleBindings and ClusterRoles, which use the `Privileged` SCC and define those SAs in the Module CR
- create RoleBindings and ClusterRoles, which use the `Privileged` SCC and use those with the default module-loader and device-plugin ServiceAccounts created by KMM
- create one or two custom SCCs (one for the module-loader and one for the device-plugin) and use those instead of the `Privileged` SCC and then also create SAs, RoleBindings, ClusterRoles as described in the 2 previous points

With this PR, KMM users have to choice to leave the afore-mentioned steps to KMM, which increases KMM's user experience. The trade-off for such improvement to UX is adding a couple of permissions to KMM itself. By following the eternal Security vs Usability dilemma, we slightly lean towards usability, without sacrificing Security to unacceptable levels. 

The end result in the PSS or SCC used in the namespace where the KMM DaemonSets will be running is exactly the same. Either the user of KMM will do the same steps. With that in mind, this PR got implemented.

@yevgeny-shnaidman @qbarrand does this provide the frame of mind for this PR?

yevgeny-shnaidman:
> > > @yevgeny-shnaidman downstream is actually the motivation for this change. We want module loader DaemonSets to be able to run out of the box - that means binding their SA to the right SCC. Right now, we have to run `oc adm policy add-scc-to-user privileged -z default` for the DaemonSet to run - this change automates that.
> > 
> > 
> > @qbarrand when i talked to @mresvanis he said that Openshift prefers that customer/user will make those changes ( just as we let the customer add labels to the namespace for pod security). In addition, if we do all this heavy lifting, why do we need the SA in the module CR?
> 
> Hey folks, let me also try to clarify, from my point of view, the motivation and trade-offs considered for this PR.
> 
> OpenShift 4.12 (Kubernetes 1.25) introduced [the enforcement of Pod Security Standards](https://connect.redhat.com/en/blog/important-openshift-changes-pod-security-standards) (via namespace labels), which lead to workloads having to adhere to 1 of the 3 standards. In addition to the PSS, workloads in OpenShift also need to adhere to a `SecurityContextConstraints`. The OpenShift [label synchronization component](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/pod-security-admission-autolabeling.md#pod-security-admission-label-synchronization) can add the respective PSS namespace labels, depending on the SCCs used in a namespace. The latter makes it a lot easier for OpenShift users to not have to deal at all with PSS, but instead keep using SCCs.
> 
> The KMM workloads (i.e. module-loader, device-plugin) belong to the `Privileged` PSS and as such, downstream KMM users need to do one of the following to successfully have their KMM DS pod admitted:
> 
> * create ServiceAccounts, RoleBindings and ClusterRoles, which use the `Privileged` SCC and define those SAs in the Module CR
> * create RoleBindings and ClusterRoles, which use the `Privileged` SCC and use those with the default module-loader and device-plugin ServiceAccounts created by KMM
> * create one or two custom SCCs (one for the module-loader and one for the device-plugin) and use those instead of the `Privileged` SCC and then also create SAs, RoleBindings, ClusterRoles as described in the 2 previous points
> 
> With this PR, KMM users have to choice to leave the afore-mentioned steps to KMM, which increases KMM's user experience. The trade-off for such improvement to UX is adding a couple of permissions to KMM itself. By following the eternal Security vs Usability dilemma, we slightly lean towards usability, without sacrificing Security to unacceptable levels.
> 
> The end result in the PSS or SCC used in the namespace where the KMM DaemonSets will be running is exactly the same. Either the user or KMM will do the same steps. With that in mind, this PR got implemented.
> 
> @yevgeny-shnaidman @qbarrand does this provide the frame of mind for this PR?
> 
> PS. @yevgeny-shnaidman `Openshift prefers that customer/user will make those changes`, OpenShift doesn't prefer this, but I'm confident that if we ask any Security team in OpenShift, they would probably tell us it's better to remove this automation. But we need to run privileged workloads either way, so we could convince them that this trade-off (between Security vs Usability) makes sense.

@mresvanis Thanks! Great explanation!  then i have no further objections to this PR, but i think we need to discuss if we really need SA in the module CR. If we have a complete automation, then having additional field will just cause confusion ( which it already did). I will open a Jira on it

mresvanis:
> @mresvanis Thanks! Great explanation! then i have no further objections to this PR, but i think we need to discuss if we really need SA in the module CR. If we have a complete automation, then having additional field will just cause confusion ( which it already did). I will open a Jira on it

Thank you, it makes sense to discuss this further and see how we can improve it even more :+1:

ybettan:
@mresvanis Can you please squash the PR into a single commit? Please see https://github.com/ybettan/kernel-module-management-midstream/blob/main/CONTRIBUTING.md.

The job to enforce it is not set in this repo yet (set only u/s ATM).

ybettan:
What does it means regarding u/s KMM? Seems like this change is triggered from the k8s security change. Don't we need to have this change u/s as well? If we do, do we have a customer urgent need for it? If not, we should probably have this change u/s first no?!

WDYT?

mresvanis:
> What does it means regarding u/s KMM? Seems like this change is triggered from the k8s security change. Don't we need to have this change u/s as well? If we do, do we have a customer urgent need for it? If not, we should probably have this change u/s first no?!
> 
> WDYT?

@ybettan [SecurityContextConstraints](https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html) is an OpenShift feature. Regarding upstream and Pod Security Standards, we would need to come up with a different solution, if we would like to offer such an automation (but I'm guessing that it would require KMM to have quite elevated permissions, which would most probably not be acceptable).

ybettan:
Talked to @mresvanis offline.
/lgtm

PR #58, opened 2022-10-02
Expand registry package API (#133)

yevgeny-shnaidman:
This PR includes:
1) adding VerifyModuleExistsForKernel function, that verifies
   presence of a file (full path) in a v1.Layer (layer of container image)
2) removing ExtractToolkitRelease funtion: no longer used
   by the preflight code, replaced by the VerifyModuleExistsForKernel
3) add unit-test
4) add test infrastructure to create a dummy v1.Layer

### DIFF
--- a/internal/registry/mock_registry_api.go
+++ b/internal/registry/mock_registry_api.go
@@ -37,21 +37,6 @@ func (m *MockRegistry) EXPECT() *MockRegistryMockRecorder {
 	return m.recorder
 }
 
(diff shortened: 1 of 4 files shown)


yevgeny-shnaidman:
Upstream PR:
https://github.com/kubernetes-sigs/kernel-module-management/commit/8dbafe605d88fe58f81010a3c87712f70519bad9

PR #60, opened 2022-10-02
Add preflight controller implementation (#158)

yevgeny-shnaidman:
This PR implements the preflight controller logic
and filtering

1) add filter function for getting preflights
   on modules change
2) add controller implementation
3) unit-test


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/8f828a76a1f0ec75b215c1da82d27ef053ab0f95

PR #61, opened 2022-10-03
Moving authenticator creation to auth package (#42)

yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/b08a55400dd8c42db786a067fac3f8cf8492a845

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/authn/kubernetes"
+	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/constants"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	k8s "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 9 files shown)


internal/auth/auth.go
```diff
@@ -59,27 +61,45 @@ func (sarag *serviceAccountRegistryAuthGetter) GetKeyChain(ctx context.Context)
...
+	if mod.Spec.ImageRepoSecret != nil {
+		namespacedName := types.NamespacedName{
+			Name:      mod.Spec.ImageRepoSecret.Name,
+			Namespace: mod.Namespace,
+		}
+		return af.newRegistryAuthGetter(namespacedName)
+	}
+	return af.newServiceAccountRegistryAuthGetter(
+		mod.Namespace,
+		constants.OCPBuilderServiceAccountName)
```
ybettan, line 104:
```suggestion
	return af.newServiceAccountRegistryAuthGetter(
		mod.Namespace,
		constants.OCPBuilderServiceAccountName
		)
```

internal/preflight/preflight.go
```diff
@@ -68,15 +67,7 @@ func (p *preflight) verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelM
...
-	var registryAuthGetter auth.RegistryAuthGetter
-	if mod.Spec.ImageRepoSecret != nil {
-		namespacedName := types.NamespacedName{
-			Name:      mod.Spec.ImageRepoSecret.Name,
-			Namespace: mod.Namespace,
-		}
-		registryAuthGetter = p.authFactory.NewRegistryAuthGetter(p.client, namespacedName)
-	}
-
+	registryAuthGetter := p.authFactory.NewRegistryAuthGetterFrom(mod)
```
ybettan, line 70: There is a change in logic here. Is it the behavior we want?

Before:
We were creating a `RegistryAuthGetter` if there were some `ImageRepoSecret` in the mod otherwise we were accessing the registry without auth.

Now:
We are creating a `RegistryAuthGetter` if there were some `ImageRepoSecret` in the mod otherwise we are creating a `ServiceAccountRegistryAuthGetter` that is not being used.

internal/auth/auth.go
yevgeny-shnaidman: i run both linter and go fmt, and it was happy with the code. Where is the convention for moving the brace to the next line?

internal/registry/registry.go
```diff
@@ -142,13 +135,17 @@ func (r *registry) getPullOptions(ctx context.Context, image string, po *kmmv1be
 	return &RepoPullConfig{repo: repo, authOptions: options}, nil
 }
 
-func (r *registry) getImageManifest(ctx context.Context, image string, pullConfig *RepoPullConfig) ([]byte, error) {
+func (r *registry) getImageManifest(ctx context.Context, image string, po *kmmv1beta1.PullOptions, registryAuthGetter auth.RegistryAuthGetter) ([]byte, *RepoPullConfig, error) {
```
ybettan, line 138: I kind of liked the fact that we are detaching the "transport" options from the other parameters. Looks cleaner to me.

internal/preflight/preflight.go
yevgeny-shnaidman: The previous code was cherry-picked from upstream. In downstream the logic for authentication is different than in upstream, and this PR synchronizes the preflight authentication with downstream

internal/auth/auth.go
ybettan: I though that this is better, but `go fmt` is indeed fixing it.
Let's keep it that way.

internal/preflight/preflight.go
ybettan: How will you access a public registry? Do we want to force the user to specify `imageRepoSecret` or `serviceAccount` in the CR even if he doesn't need it?

internal/registry/registry.go
ybettan: I probably missed it u/s :( 
I think it complicates it but it is still better to keep the sync with u/s.
We can always change it u/s later.

internal/preflight/preflight.go
yevgeny-shnaidman: we don't force user. In  case imageRepoSecret is not specified, the code will use the build ServiceAccount defined for each namespace by OCP. This logic is exactly the same that is already implemented in checkImageExists

ybettan: So it that case, if the user specify a serviceAccount in the CR, we will ignore it because we are taking the default service account?

yevgeny-shnaidman: yes. user's SA is for security reasons, not for builder

ybettan: Ack.

PR #70, opened 2022-10-06
Preflight should use PullOptions in image verification (#52)

yevgeny-shnaidman:
Since now we have pull options also in kernel mapping and in Spec, those parameters should now be used in the preflight image verification


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/809e889ec87aa30dfa197cad805e815e3e687d52

PR #74, opened 2022-10-11
Change all SRO references to KMM/Modules

enriquebelarte:
Change and substitute old references to Special Resource Operator to KMM/Modules.


qbarrand:
Would you mind opening this upstream instead?
/hold

PR #80, opened 2022-10-18
Update of Build configuration in Module should trigger new BuildConfig

yevgeny-shnaidman:
Currently once the BuildConfig is created, it can be changed (edited) but it won't affect a currently running Build and its pod. This fixes it by doing the following
1) when BuildConfig is created, BuildSource data is hashed and saved as annotation
   in the BuildConfig
2) when checking BuildConfig status, hash of the new contents is compared
   to the existing one
3) in case hashes differ, the current BuildConfig is deleted, which in turn deletes
   its Build and pod, and then new BuildConfig will be created in the next reconcilation loop

### DIFF
--- a/internal/build/buildconfig/maker.go
+++ b/internal/build/buildconfig/maker.go
@@ -3,6 +3,7 @@ package buildconfig
 import (
 	"fmt"
 
+	"github.com/mitchellh/hashstructure"
 	buildv1 "github.com/openshift/api/build/v1"
 	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/build"
@@ -17,7 +18,7 @@ import (
 //go:generate mockgen -source=maker.go -package=buildconfig -destination=mock_maker.go
 
 type Maker interface {
-	MakeBuildConfig(mod kmmv1beta1.Module, mapping kmmv1beta1.KernelMapping, targetKernel, containerImage string, pushImage bool) (*buildv1.BuildConfig, error)
+	MakeBuildConfigTemplate(mod kmmv1beta1.Module, mapping kmmv1beta1.KernelMapping, targetKernel, containerImage string, pushImage bool) (*buildv1.BuildConfig, error)
 }
 
 type maker struct {
@@ -32,7 +33,7 @@ func NewMaker(helper build.Helper, scheme *runtime.Scheme) Maker {
 	}
 }
 
-func (m *maker) MakeBuildConfig(mod kmmv1beta1.Module, mapping kmmv1beta1.KernelMapping, targetKernel, containerImage string, pushImage bool) (*buildv1.BuildConfig, error) {
+func (m *maker) MakeBuildConfigTemplate(mod kmmv1beta1.Module, mapping kmmv1beta1.KernelMapping, targetKernel, containerImage string, pushImage bool) (*buildv1.BuildConfig, error) {
 	kmmBuild := m.helper.GetRelevantBuild(mod, mapping)
 
 	buildArgs := m.helper.ApplyBuildArgOverrides(
@@ -51,11 +52,22 @@ func (m *maker) MakeBuildConfig(mod kmmv1beta1.Module, mapping kmmv1beta1.Kernel
 		buildTarget = buildv1.BuildOutput{}
 	}
 
+	sourceConfig := buildv1.BuildSource{
+		Dockerfile: &kmmBuild.Dockerfile,
+		Type:       buildv1.BuildSourceDockerfile,
+	}
+
+	sourceConfigHash, err := hashstructure.Hash(sourceConfig, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not hash BuildConfig's Buildsource template: %v", err)
+	}
+
 	bc := buildv1.BuildConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: mod.Name + "-",
 			Namespace:    mod.Namespace,
 			Labels:       build.GetBuildLabels(mod, targetKernel),
+			Annotations:  map[string]string{buildConfigHashAnnotation: fmt.Sprintf("%d", sourceConfigHash)},
 		},
 		Spec: buildv1.BuildConfigSpec{
 			Triggers: []buildv1.BuildTriggerPolicy{
@@ -64,10 +76,7 @@ func (m *maker) MakeBuildConfig(mod kmmv1beta1.Module, mapping kmmv1beta1.Kernel
 			RunPolicy: buildv1.BuildRunPolicySerialLatestOnly,
 			CommonSpec: buildv1.CommonSpec{
 				ServiceAccount: constants.OCPBuilderServiceAccountName,
-				Source: buildv1.BuildSource{
-					Dockerfile: &kmmBuild.Dockerfile,
-					Type:       buildv1.BuildSourceDockerfile,
-				},
+				Source:         sourceConfig,
 				Strategy: buildv1.BuildStrategy{
 					Type: buildv1.DockerBuildStrategyType,
 					DockerStrategy: &buildv1.DockerBuildStrategy{
--- a/internal/build/buildconfig/manager.go
+++ b/internal/build/buildconfig/manager.go
@@ -8,9 +8,12 @@ import (
 	buildv1 "github.com/openshift/api/build/v1"
 	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/build"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/utils"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
+
+const buildConfigHashAnnotation = "kmm.node.kubernetes.io/last-hash"
 
 var (
 	ErrNoMatchingBuildConfig = errors.New("no matching BuildConfig")
@@ -34,6 +37,11 @@ func NewManager(client client.Client, maker Maker, ocpBuildsHelper OpenShiftBuil
 func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, pushImage bool) (build.Result, error) {
 	logger := log.FromContext(ctx)
 
+	buildConfigTemplate, err := bcm.maker.MakeBuildConfigTemplate(mod, m, targetKernel, m.ContainerImage, pushImage)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not make BuildConfig template: %v", err)
+	}
+
 	buildConfig, err := bcm.ocpBuildsHelper.GetBuildConfig(ctx, mod, targetKernel)
 	if err != nil {
 		if !errors.Is(err, ErrNoMatchingBuildConfig) {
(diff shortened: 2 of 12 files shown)


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/6171fb7b56dea37ac7604e8881075338a27e3dd5

ybettan:
1. Isn't OCP `BuildConfig` changes create/updates the builds?
2. If it doesn't, have you checked if you can add an annotation to the `BuildConfig` to make OCP update the builds on `BuildConifig`s updates? (I know there are such for rebooting pods using `Secrets`/`Configmaps` when those are changed.

yevgeny-shnaidman:
> 1. Isn't OCP `BuildConfig` changes create/updates the builds?
> 2. If it doesn't, have you checked if you can add an annotation to the `BuildConfig` to make OCP update the builds on `BuildConifig`s updates? (I know there are such for rebooting pods using `Secrets`/`Configmaps` when those are changed.

BuildConfig is supposed to create/update builds upon editing, but i found a document where it says that it will be supported later. The document is from 2 years ago, but i checked it on 4.12 and it still  does not support it. I found no other way to cause the change to Build/Pod. I think for now we can work the same way as we do in upstream. Once BuildConfig starts supporting editing, we can change the code

internal/build/buildconfig/maker.go
```diff
@@ -51,11 +52,22 @@ func (m *maker) MakeBuildConfig(mod kmmv1beta1.Module, mapping kmmv1beta1.Kernel
 		buildTarget = buildv1.BuildOutput{}
 	}
 
+	sourceConfig := buildv1.BuildSource{
+		Dockerfile: &kmmBuild.Dockerfile,
+		Type:       buildv1.BuildSourceDockerfile,
+	}
```
ybettan, line 58: So it means that if the user has changed some data in the BuildConfig which is not the Dockerfile itself, the build won't be retrigerred?!

Don't we want to hash the full buildConfig?

internal/build/buildconfig/manager.go
```diff
@@ -34,6 +37,11 @@ func NewManager(client client.Client, maker Maker, ocpBuildsHelper OpenShiftBuil
 func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, pushImage bool) (build.Result, error) {
 	logger := log.FromContext(ctx)
 
+	buildConfigTemplate, err := bcm.maker.MakeBuildConfigTemplate(mod, m, targetKernel, m.ContainerImage, pushImage)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not make BuildConfig template: %v", err)
```
ybettan, line 42:
```suggestion
		return build.Result{}, fmt.Errorf("could not make BuildConfig template: %w", err)
```

internal/build/buildconfig/manager.go
```diff
@@ -42,18 +50,27 @@ func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module,
...
 
-		buildConfig, err = bcm.maker.MakeBuildConfig(mod, m, targetKernel, m.ContainerImage, pushImage)
-		if err != nil {
-			return build.Result{}, fmt.Errorf("could not make BuildConfig: %v", err)
-		}
-
-		if err = bcm.client.Create(ctx, buildConfig); err != nil {
-			return build.Result{}, fmt.Errorf("could not create Job: %v", err)
+		if err = bcm.client.Create(ctx, buildConfigTemplate); err != nil {
+			return build.Result{}, fmt.Errorf("could not create BuildConfig: %v", err)
```
ybettan, line 54:
```suggestion
			return build.Result{}, fmt.Errorf("could not create BuildConfig: %w", err)
```

internal/build/buildconfig/maker.go
yevgeny-shnaidman: The data that user can change ( in the Module spec) is only used in the BuildSource. All the rest are set by operator code. So, in case of Module change, we only need to compare hashes of the BuildSource

internal/build/buildconfig/manager.go
```diff
@@ -42,18 +50,27 @@ func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module,
...
 		}
 
 		return build.Result{Status: build.StatusCreated, Requeue: true}, nil
 	}
 
+	changed, err := bcm.isBuildConfigChanged(buildConfig, buildConfigTemplate)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not determine if BuildConfig has changed: %v", err)
```
ybettan, line 62:
```suggestion
		return build.Result{}, fmt.Errorf("could not determine if BuildConfig has changed: %w", err)
```

internal/build/buildconfig/manager.go
```diff
@@ -42,18 +50,27 @@ func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module,
...
+	}
+
+	if changed {
+		logger.Info("The module's build spec has been changed, deleting the current BuildConfig so a new one can be created", "name", buildConfig.Name)
```
ybettan, line 66:
```suggestion
		logger.Info("The module's build spec has been changed, deleting the current Build resource so a new one can be created", "name", buildConfig.Name)
```

internal/build/buildconfig/manager.go
yevgeny-shnaidman: If i remember correctly, we agreed that we will be using %v

internal/build/buildconfig/manager.go
```diff
@@ -42,18 +50,27 @@ func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module,
...
+		err = bcm.client.Delete(ctx, buildConfig)
```
ybettan, line 67: Don't we only need to delete the build and let the buildConfig generate a new one?

internal/build/buildconfig/manager.go
yevgeny-shnaidman: same comment as above

internal/build/buildconfig/manager.go
yevgeny-shnaidman: same comment as above

internal/build/buildconfig/manager.go
yevgeny-shnaidman: i don't see any significant advantage in deleting the Build, and not BuildConfig:
1) deleting the BuildConfig makes the code simplier, since we already have that information, and don't need to to look for relevant Build
2) if we delete Build, then we need to edit BuildConfig first. This means additional code. In case we delete BuildConfig, no additional code for editing needed, since the code will simply create new BuildConfig in the next reconciliation loop

internal/build/buildconfig/maker.go
ybettan: Makes perfect sense. Thanks.

internal/build/buildconfig/manager.go
ybettan: Did we? I remember we said we want to wrap errors (using %w). Also I see we are using it extensively:
```
root kernel-module-management-midstream (main) $ git grep %w | grep -v vendor | grep %w
controllers/module_reconciler.go:               return res, fmt.Errorf("failed to get the requested %s KMMO CR: %w", req.NamespacedName, err)
controllers/module_reconciler.go:               return res, fmt.Errorf("could get targeted nodes for module %s: %w", mod.Name, err)
controllers/module_reconciler.go:               return res, fmt.Errorf("could get kernel mappings and nodes for modules %s: %w", mod.Name, err)
controllers/module_reconciler.go:                       return res, fmt.Errorf("failed to handle build for kernel version %s: %w", kernelVersion, err)
controllers/module_reconciler.go:               return res, fmt.Errorf("could handle device plugin: %w", err)
controllers/module_reconciler.go:               return res, fmt.Errorf("failed to update status of the module: %w", err)
controllers/module_reconciler.go:               return false, fmt.Errorf("failed to check image existence for kernel %s: %w", kernelVersion, err)
controllers/module_reconciler.go:               return false, fmt.Errorf("could not synchronize the build: %w", err)
controllers/module_reconciler.go:               return fmt.Errorf("failed to get the device plugin daemonset %s/%s: %w", name, mod.Namespace, err)
controllers/module_reconciler.go:               return nil, fmt.Errorf("failed to get the kmmo module %s: %w", namespacedName, err)
internal/auth/auth.go:          return nil, fmt.Errorf("cannot find secret %s: %w", rsag.namespacedName, err)
internal/auth/auth.go:          return nil, fmt.Errorf("could not create a keycahin from secret %v: %w", secret, err)
internal/daemonset/daemonset.go:                return nil, fmt.Errorf("could not get all DaemonSets: %w", err)
internal/module/kernelmapper.go:                return nil, fmt.Errorf("failed to substitute the os config into ContainerImage field: %w", err)
internal/registry/registry.go:          return false, fmt.Errorf("failed to get pull options for image %s: %w", image, err)
internal/registry/registry.go:          return false, fmt.Errorf("could not get image %s: %w", image, err)
internal/registry/registry.go:          return nil, nil, fmt.Errorf("failed to get pull options for image %s: %w", image, err)
internal/registry/registry.go:          return nil, nil, fmt.Errorf("failed to get manifest from image %s: %w", image, err)
internal/registry/registry.go:          return nil, nil, fmt.Errorf("failed to get layers digests from manifest of the image %s: %w", image, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to find file %s in image layer: %w", driverToolkitJSONFile, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get KERNEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get RT_KERNEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get RHEL_VERSION from %s, found %t: %w", driverToolkitJSONFile, found, err)
internal/registry/registry.go:                  return nil, fmt.Errorf("cannot get keychain from the registry auth getter: %w", err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get manifest stream from image %s: %w", image, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get crane manifest from image %s: %w", image, err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to unmarshal crane manifest: %w", err)
internal/registry/registry.go:          return nil, fmt.Errorf("unmarshalled manifests invalid format: %w", err)
internal/registry/registry.go:                  return nil, fmt.Errorf("failed to get arch digets from multi arch image: %w", err)
internal/registry/registry.go:                  return nil, fmt.Errorf("failed to get crane manifest for the arch image: %w", err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to unmarshal manifest stream: %w", err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to get targz from layer: %w", err)
internal/registry/registry.go:          return nil, fmt.Errorf("failed to create reader from targz: %w", err)
internal/registry/registry.go:                  return nil, fmt.Errorf("failed to get next entry from targz: %w", err)
internal/registry/registry.go:                          return nil, fmt.Errorf("failed to read tar entry: %w", err)
internal/registry/registry.go:                          return nil, fmt.Errorf("failed to unmarshal tar entry: %w", err)
internal/registry/registry.go:          return "", fmt.Errorf("failed to unmarshal manifest stream: %w", err)
```

yevgeny-shnaidman: we are also using %v extensivily

internal/build/buildconfig/manager.go
ybettan: Ack.

internal/build/buildconfig/manager.go
ybettan: Yes, sorry, I though that we are only deleting the build. You can ignore that comment.

PR #83, opened 2022-10-27
Non-schedulable nodes should not be part of targeted nodes for Module…

yevgeny-shnaidman:
… (#125)

Currently ModuleLoader cannot be scheduled to run on Non-Schedulable nodes, i.e. masters etc'. This PR fixed the extraction of schedulable nodes list, which will also affect the status update of the Module

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -241,13 +241,20 @@ func (r *ModuleReconciler) getNodesListBySelector(ctx context.Context, mod *kmmv
 	logger := log.FromContext(ctx)
 	logger.V(1).Info("Listing nodes", "selector", mod.Spec.Selector)
 
-	nodes := v1.NodeList{}
+	selectedNodes := v1.NodeList{}
 	opt := client.MatchingLabels(mod.Spec.Selector)
-	if err := r.Client.List(ctx, &nodes, opt); err != nil {
+	if err := r.Client.List(ctx, &selectedNodes, opt); err != nil {
 		logger.Error(err, "Could not list nodes")
 		return nil, fmt.Errorf("could not list nodes: %v", err)
 	}
-	return nodes.Items, nil
+	nodes := make([]v1.Node, 0, len(selectedNodes.Items))
+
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/c26cefd073c238b4fe0bf85a814594d3cdbee0b4

controllers/module_reconciler.go
```diff
@@ -241,13 +241,20 @@ func (r *ModuleReconciler) getNodesListBySelector(ctx context.Context, mod *kmmv
...
-	nodes := v1.NodeList{}
+	selectedNodes := v1.NodeList{}
 	opt := client.MatchingLabels(mod.Spec.Selector)
-	if err := r.Client.List(ctx, &nodes, opt); err != nil {
+	if err := r.Client.List(ctx, &selectedNodes, opt); err != nil {
 		logger.Error(err, "Could not list nodes")
 		return nil, fmt.Errorf("could not list nodes: %v", err)
 	}
-	return nodes.Items, nil
+	nodes := make([]v1.Node, 0, len(selectedNodes.Items))
```
ybettan, line 250: I would prefer to call it `scheduableNodes` but let's stay consistent with the u/s code.

controllers/module_reconciler.go
```diff
@@ -397,3 +404,12 @@ func (r *ModuleReconciler) SetupWithManager(mgr ctrl.Manager, kernelLabel string
 		Named("module").
 		Complete(r)
 }
+
+func isNodeSchedulable(node *v1.Node) bool {
+	for _, taint := range node.Spec.Taints {
```
ybettan, line 409: Who is responsible for tainting the node?

yevgeny-shnaidman: master nodes , that are not schedulable ( not SNO) are tainted by KubeAPI server

ybettan: Ack. So this has nothing to do with our code.

PR #88, opened 2022-11-03
Add check-commits-count

enriquebelarte:
Add script to check that each PR consists of one single commit.

### DIFF
--- a/ci/prow/check-commits-count
+++ b/ci/prow/check-commits-count
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+git remote add main https://github.com/rh-ecosystem-edge/kernel-module-management.git
+git fetch main
+
+commits_count=$(git rev-list --count HEAD ^main)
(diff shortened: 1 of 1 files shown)


ci/prow/check-commits-count
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+number_of_commits=$(git rev-list --count HEAD ^main)
+if [[ ${number_of_commits} != 1 ]]; then
```
ybettan, line 6: Did you test it? Why don't we check the last 2 commits like we do u/s?
https://github.com/kubernetes-sigs/kernel-module-management/pull/69/files

ybettan:
Latest update for u/s code: https://github.com/kubernetes-sigs/kernel-module-management/pull/69

ci/prow/check-commits-count
enriquebelarte: You're right. Modified.

PR #91, opened 2022-11-04
Fix kustomize and bundle CRD manifests YAML errors

mresvanis:
This PR fixes `kustomize` and `bundle` CRD manifests YAML errors leading to failed `kustomize` deployment. Specifically, it resolves the following issue:

```shell
$ oc apply -k https://github.com/rh-ecosystem-edge/kernel-module-management/config/default
error: accumulating resources: accumulation err='accumulating resources from '../crd': '/tmp/kustomize-2948061914/config/crd' must resolve to a file': recursed accumulation of path '/tmp/kustomize-2948061914/config/crd': accumulating resources: accumulating resources from 'bases/kmm.sigs.k8s.io_preflightvalidations.yaml': MalformedYAMLError: yaml: line 24: could not find expected ':' in File: bases/kmm.sigs.k8s.io_preflightvalidations.yaml
```

Signed-off-by: Michail Resvanis <mresvani@redhat.com>

PR #92, opened 2022-11-06
add the fields for signing to crd (#91)

yevgeny-shnaidman:
this add four fields to the crd in their own subsection:

FileList - a list of kmod files within the container to sign

KeySecret and CertSecret - these are the names of secrets containing the private and public keys respectivly that will be used to sign the kmods. The public key should be the one added into the uefi moklist for secureboot,
 and KeySecret is its private component. (See the kernel docs e.g.
https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#manually-signing-modules for details of this process)

unsignedImage - the optional name of a dtk image containing the unsigned knods, used only if their is no build stage.

The expected workflow will be (not implemented in this commit): if there is only a kernelmapping.build section it will produce the image kernelmapping.containerImage (for the non-secureboot case)

if there is both km.build and km.sign then build will produce an intermediate image and sign will consume that to produce km.containerImage

if there is only km.sign signing will consumse km.sign.unsignedImage and produce km.containerImage (this is for the case where a vendor supplies prebuilt images that need signing)

the generation of the intermediate image name will be handled automatically within the controller (again in a different commit) but will probably be something like km.containerImage + "-unsigned"


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/4c6a77d4f88fa7839ba20d80e081167196028f06

PR #94, opened 2022-11-06
centralise the name of the image to build in the controller (#88)

yevgeny-shnaidman:
The name of the image to build is assumed to be km.ContainerImage in several places in the code, this commit centralises that assumption into a single point in Reconcile(). In the future this will allow us to determine the image the stage produces once E.g if we add signing (or other) intermediate stages to the image build process.


yevgeny-shnaidman:
Pull cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/c4db89206f405a126d68d2b3c17acfaca436549b

PR #96, opened 2022-11-08
Adding the consumption of DTK by KMM.

ybettan:
If the user has specified the DTK_AUTO build arg in his Dockerfile, KMMO will now find the correct DTK image for each kernel version and add that build ARG to the Dockerfile.

This commit is following the design in
docs/enhancements/using-dtk-for-building-driver-containers.md


ybettan:
/cc @yevgeny-shnaidman 
/hold
I want to add 2 more UT for the updated `Reconcile` funcions.
The code should be ready though.

ybettan:
FYI @enriquebelarte

docs/enhancements/using-dtk-for-building-driver-containers.md
```diff
@@ -93,8 +93,8 @@ If the user want us to pull the correct DTK image on builds/cluster-upgrade, he
 
 #### Dockerfile which let KMMO find the correct DTK image.
 ```
-ARG DTK-AUTO
-FROM ${DTK-AUTO}
+ARG DTK_AUTO
```
yevgeny-shnaidman, line 96: are we going to support DTK only as argument to Dockerfile, or also as a template in the dockerfile that is going to be replaced ( just as template strings in images?

controllers/module_reconciler.go
```diff
@@ -280,6 +324,10 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
 
 	logger.Info("Image not pull-able; building in-cluster")
 
+	if err := r.injectDTKToBuildIfNeeded(ctx, kernelVersion, km.Build); err != nil {
```
yevgeny-shnaidman, line 327: This is a little problematic: km.Build can be nil.

yevgeny-shnaidman, line 327: WDYT about moving the injectDTKToBuildIfNeeded into build package, where it is going to be called by in the Sync API?

docs/enhancements/using-dtk-for-building-driver-containers.md
qbarrand: What would be the benefit of a template? I would stick to the build arg, which is what a Dockerfile writer would expect.

controllers/globals.go
```diff
@@ -0,0 +1,5 @@
+package controllers
+
+// Currently we only have 1 written and 1 reader so there is no reason to
```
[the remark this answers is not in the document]
qbarrand: If both the reader and the writer access the map concurrently (which sounds possible?) then the code will panic. I would wrap this with a mutex now.

controllers/module_reconciler.go
```diff
@@ -257,6 +262,45 @@ func (r *ModuleReconciler) getNodesListBySelector(ctx context.Context, mod *kmmv
 	return nodes, nil
 }
 
+func (r *ModuleReconciler) injectDTKToBuildIfNeeded(ctx context.Context, kernelVersion string, build *kmmv1beta1.Build) error {
+
+	logger := log.FromContext(ctx, "kernel version")
+
+	if dockerfile := build.Dockerfile; strings.Contains(dockerfile, dtkBuildArg) {
```
qbarrand, line 269: I would always inject the build argument and not call `strings.Contains` here.

controllers/node_kernel_reconciler.go
```diff
@@ -37,7 +38,14 @@ func (r *NodeKernelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, fmt.Errorf("could not get node: %v", err)
 	}
 
+	// We expect an osVersion of the form 411.86.202210072320-0 for example
+	osVersionRegexp := regexp.MustCompile("[0-9]+.[0-9]+.[0-9]+-[0-9]")
```
qbarrand, line 42: `.` will match any character, not just dots.
```suggestion
	osVersionRegexp := regexp.MustCompile(`\d+\.\d+\.\d+\-\d`)
```
We should also move this outside of the function so it's ran only once.

controllers/globals.go
```diff
@@ -0,0 +1,5 @@
...
+// add mutexts. If writters are added, we need to make sure to add syncronization.
+var kernelToOS = map[string]string{}
```
qbarrand, line 5: I think it would be more convenient to map the kernel version to the DTK image here.
This likely means that we need to watch both nodes and the DTK ImageStream to react on updates on both. Sounds like a good reason to move the new code from the node-kernel reconciler to a separate one, IMHO.

controllers/globals.go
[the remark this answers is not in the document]
yevgeny-shnaidman: @qbarrand why would code panic? at the worst case, etiher the reader won't find the key ( it's ok, will find it in the next iteration), or the read value will be nil value, which can easily be checked.

controllers/globals.go
yevgeny-shnaidman: @qbarrand why do we need to watch the image stream? We can get its data once we really need to process the build.

controllers/globals.go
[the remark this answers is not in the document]
qbarrand: @yevgeny-shnaidman https://go.dev/play/p/f9Klsm03D8C

controllers/globals.go
qbarrand: If we watch it, we only have to find the image name for a given OS version once - vs each time we need to create a build.

yevgeny-shnaidman: that can easily be mitigated with caching, instead of additional controller...

controllers/globals.go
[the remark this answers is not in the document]
yevgeny-shnaidman: i see, Golang is not C , unfortunately. Ok, let's add ReadWriter lock

controllers/globals.go
qbarrand: I don't see any downside in having that feature in another controller - clear separation of concerns. The Node informer would be shared together with its cache, so the impact on memory utilization would be limited to the ImageStream watcher.

controllers/globals.go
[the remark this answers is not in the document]
ybettan: I was not aware of that either. Will add a lock then.

internal/filter/filter.go
```diff
@@ -46,12 +46,29 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			kernelVersionChanged := e.ObjectNew.GetLabels()[labelName] != e.ObjectNew.(*v1.Node).Status.NodeInfo.KernelVersion
+			osImageVersionChanged := e.ObjectNew.(*v1.Node).Status.NodeInfo.OSImage != e.ObjectOld.(*v1.Node).Status.NodeInfo.OSImage
+			return kernelVersionChanged || osImageVersionChanged
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return e.Object.GetLabels()[labelName] != e.Object.(*v1.Node).Status.NodeInfo.KernelVersion
```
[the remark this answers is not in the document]
ybettan: This new filter is combining the filter of the "map reconciliation"  and the "label reconciliation".
The previous implementation was using the generic event so I added it only to keep backward compatibility.

If you don't think it matters, I don't mind removing it.

docs/enhancements/using-dtk-for-building-driver-containers.md
ybettan: It looks clean to use the build arg as you can inspect the buildConfig, see that `DTK_AUTO` specified by the user and we only added a build arg (instead of manipulating the user's Dockerfile).

Closing this thread.

controllers/module_reconciler.go
ybettan: Will add a validation to skip `nil`.
Thanks.

controllers/globals.go
ybettan: To me, the real value of mapping `kernel-->DTK` is that it will simplify the hub-spoke flow as if a spoke send its kernels and DTKs (instead of kernel+osImages) then we don't need to worry about that specific ocImages not being present in the hub's imagestream.

internal/filter/filter.go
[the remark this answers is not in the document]
yevgeny-shnaidman: lets remove it, it will be safer

controllers/module_reconciler.go
yevgeny-shnaidman: you need more then checking for nil: you need to decide what Build to use, the one in the mapping or the global one. there is a function that does just that: GetRelevantBuild, and it is called at the beginning of Sync function. This is one of the reasons that i asked you to consider whether the inject function needs to be executed in the Sync, since it is part of creating the Build

internal/filter/filter.go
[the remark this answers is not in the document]
ybettan: Removed.

controllers/module_reconciler.go
ybettan: IMO. A clear code takes priority to a "fast code" until a bottleneck is created.
I think it is wired (and harder to understand when debugging) to check the buildconfig, see a build arg that doesn't exist in the Dockerfile.

I would stick with adding that build arg only if the user is using the `DTK_AUTO` feature.

qbarrand: @yevgeny-shnaidman WDYT? should we always add the build arg or only when we find the substring in the Dockerfile?

yevgeny-shnaidman: I personally prefer to go Yoni's way: we are not writing RT application, so strings.Contains performance is negligible in this case, but it does makes code clearer and self-explanatory

controllers/globals.go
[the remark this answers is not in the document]
ybettan: Done. Added a simple lock for simplicity. We can use 2 locks (one for each map) but it seemed redundant and make the code more complex and since node/imagestream events are rare then I don't see the justification for it.

controllers/module_reconciler.go
ybettan: Done.

ybettan: Done.

ybettan:
Still need to update the UT but the code is 95% done.
@yevgeny-shnaidman, @qbarrand PTAL.

controllers/node_kernel_reconciler.go
```diff
@@ -14,6 +15,9 @@ import (
 
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;patch;list;watch
 
+// We expect an osImage of the form 411.86.202210072320-0 for example
+var osVersionRegexp = regexp.MustCompile(`\d+\.\d+\.\d+\-\d`)
```
qbarrand, line 19: 🙇🏻‍♂️

controllers/globals.go
```diff
@@ -0,0 +1,5 @@
+package controllers
+
+import "github.com/rh-ecosystem-edge/kernel-module-management/internal/syncronizedmap"
+
+var syncronizedKernelOsDtkMapping = syncronizedmap.NewSyncronizedKernelOsDtkMapping()
```
qbarrand, line 5: I think this should be an instance variable of all 3 reconcilers.

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,54 @@
...
+
+//FIXME: add UT
+type SyncronizedKernelOsDtkMapping interface {
+	NewNodeInfo(kernelVersion, osImage string)
+	NewImageStreamInfo(osImage, dtkImage string)
+	GetImage(kernelVersion string) (string, bool)
+}
+
+type syncronizedKernelOsDtkMapping struct {
+	m          *sync.RWMutex
```
qbarrand, line 13: IMHO, no need to block writes into `osToDtk` when we are writing to `kernelToOs`. I would have two mutexes here.
Or maybe only one map? WDYT?

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,54 @@
...
+	kernelToOs map[string]string
+	osToDtk    map[string]string
```
qbarrand, line 15: Capitalized words can remain capital in variable names:
```suggestion
	kernelToOS map[string]string
	osToDTK    map[string]string
```

controllers/globals.go
ybettan: https://github.com/rh-ecosystem-edge/kernel-module-management/pull/96#discussion_r1020243279
This is what I think :)

internal/syncronizedmap/syncronizedmap.go
ybettan: https://github.com/rh-ecosystem-edge/kernel-module-management/pull/96#discussion_r1020243279
This is what I think :)

controllers/globals.go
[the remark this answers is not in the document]
qbarrand: How is it redundant? Even though we do not expect a high rate of updates, blocking access to a data structure we're not touching doesn't sound right.

controllers/globals.go
yevgeny-shnaidman: i agree with @qbarrand : this can be initialized in the main and passed as parameter during reconcilers'initialization

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (skom *syncronizedKernelOsDtkMapping) GetImage(kernelVersion string) (string, bool) {
+
+	skom.m.RLock()
+	defer skom.m.RUnlock()
+
+	osImage, ok := skom.kernelToOs[kernelVersion]
+	if !ok {
+		return "", false
```
yevgeny-shnaidman, line 50: lets add an Error log here, since this function does not return error

internal/build/buildconfig/maker.go
```diff
@@ -33,12 +39,28 @@ func NewMaker(helper build.Helper, scheme *runtime.Scheme) Maker {
...
+		return nil, fmt.Errorf("could not get DTK image for kernel %v", targetKernel)
+	}
+
+	overrides := []kmmv1beta1.BuildArg{
+		{
+			Name:  "KERNEL_VERSION",
+			Value: targetKernel,
+		},
+	}
+	if dockerfile := kmmBuild.Dockerfile; strings.Contains(dockerfile, dtkBuildArg) {
```
yevgeny-shnaidman, line 58:
```suggestion
	if strings.Contains(dockerfile, dtkBuildArg) {
```

controllers/node_kernel_reconciler.go
```diff
@@ -38,6 +42,11 @@ func (r *NodeKernelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	kernelVersion := node.Status.NodeInfo.KernelVersion
+	osImage := osVersionRegexp.FindString(node.Status.NodeInfo.OSImage)
+	if osImage == "" {
+		return ctrl.Result{}, fmt.Errorf("could not get node %s osImage", node.Name)
+	}
+	syncronizedKernelOsDtkMapping.NewNodeInfo(kernelVersion, osImage)
```
yevgeny-shnaidman, line 49: what about changing the predicate for this reconciler?

internal/syncronizedmap/syncronizedmap.go
ybettan: I though we don't want any error log in inner packages..

controllers/node_kernel_reconciler.go
ybettan: Good catch. Forgot about it.

internal/build/buildconfig/maker.go
ybettan: I am not sure what you meant here. `dockerfile` isn't defined if you remove the variable definition.

yevgeny-shnaidman: sorry, bug in suggestion:

if strings.Contains(kmmBuild.dockerfile, dtkBuildArg) {

internal/syncronizedmap/syncronizedmap.go
yevgeny-shnaidman: in general we don't, since we stack them all in the returned errors. But in this case the error is not returned, so we do need some log in order to debug it

qbarrand: Or we change this method to return an error?

controllers/globals.go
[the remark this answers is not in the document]
ybettan: Added a second mutex.

internal/syncronizedmap/syncronizedmap.go
ybettan: I think it is clearer that way. If I change it, I will also change it in other places and sometimes is is more confusing than helping.

For exampel:
`SyncronizedKernelOsDtkMapping` vs `SyncronizedKernelOSDTKMapping`

internal/syncronizedmap/syncronizedmap.go
ybettan: I prefer to keep the behavior as close as possible to a golang map access.
The caller of this method (`MakeBuildConfigTemplate`) will return a valid error in case this method return `false`.

internal/build/buildconfig/maker.go
ybettan: You are right. I did it cause I used `dockerfile` in the loop but I changed it and forgot to remove it from the `if` statement as well.

internal/syncronizedmap/syncronizedmap.go
yevgeny-shnaidman: you have 2 options here to return false: kernel not present in the map or os not present in the map. Adding a log here will help to pinpoint the issue in case of an error

controllers/node_kernel_reconciler.go
ybettan: Done.

ybettan: Done.

internal/syncronizedmap/syncronizedmap.go
ybettan: As agreed with @yevgeny-shnaidman we will leave it as is.

controllers/globals.go
ybettan: Fixed.

controllers/module_reconciler_test.go
```diff
@@ -703,6 +713,7 @@ var _ = Describe("ModuleReconciler_handleBuild", func() {
 		mockMetrics     *metrics.MockMetrics
 		mockSU          *statusupdater.MockModuleStatusUpdater
 		mockRegistry    *registry.MockRegistry
+		//mockSKODM       *syncronizedmap.MockSyncronizedKernelOsDtkMapping //FIXME: do we need it?
```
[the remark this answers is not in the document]
ybettan: Not yet. Working on the UT.

controllers/module_reconciler_test.go
```diff
@@ -964,12 +981,57 @@ var _ = Describe("ModuleReconciler_handleBuild", func() {
...
 			mockSU,
+			nil,
 		)
 
 		res, err := mr.handleBuild(context.Background(), mod, km, kernelVersion, km.ContainerImage)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(res).To(BeFalse())
 	})
+
+	//FIXME: this should move to another API
```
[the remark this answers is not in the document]
ybettan: Not yet. Working on the UT.

controllers/module_reconciler_test.go
```diff
@@ -716,6 +727,7 @@ var _ = Describe("ModuleReconciler_handleBuild", func() {
 		mockMetrics = metrics.NewMockMetrics(ctrl)
 		mockSU = statusupdater.NewMockModuleStatusUpdater(ctrl)
 		mockRegistry = registry.NewMockRegistry(ctrl)
+		//mockSKODM = syncronizedmap.NewMockSyncronizedKernelOsDtkMapping //FIXME: do we need it?
```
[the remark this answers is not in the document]
ybettan: Not yet. Working on the UT.

controllers/module_reconciler_test.go
```diff
@@ -1037,3 +1099,125 @@ var _ = Describe("ModuleReconciler_getNodesListBySelector", func() {
 		Expect(len(nodeList)).To(Equal(1))
 	})
 })
+
+////FIXME: uncomment
```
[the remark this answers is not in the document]
ybettan: Not yet. Working on the UT.

controllers/imagestream_reconciler.go
```diff
@@ -0,0 +1,58 @@
...
+
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/filter"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/syncronizedmap"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//+kubebuilder:rbac:groups="core",resources=imagestreams,verbs=get;patch;list;watch
```
qbarrand, line 15: I don't think we are patching the ImageStream?
```suggestion
//+kubebuilder:rbac:groups="core",resources=imagestreams,verbs=get;list;watch
```

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,67 @@
+package syncronizedmap
+
+import (
+	"fmt"
+	"sync"
+)
+
+//go:generate mockgen -source=syncronizedmap.go -package=syncronizedmap -destination=mock_syncronizedmap.go
+
+type SyncronizedKernelOsDtkMapping interface {
```
qbarrand, line 10: Is `Synchronized` needed here?
```suggestion
type KernelOsDtkMap interface {
```

controllers/imagestream_reconciler.go
ybettan: Yea we don't patch, I will remove it.

internal/syncronizedmap/syncronizedmap.go
ybettan: Not really, I can remove it.

controllers/imagestream_reconciler.go
ybettan: Removed.

internal/syncronizedmap/syncronizedmap.go
ybettan: Updated

controllers/imagestream_reconciler.go
```diff
@@ -0,0 +1,58 @@
...
+
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/filter"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/syncronizedmap"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//+kubebuilder:rbac:groups="core",resources=imagestreams,verbs=get;list;watch
```
qbarrand, line 15:
```suggestion
//+kubebuilder:rbac:groups="image.openshift.io",resources=imagestreams,namespace="driver-toolkit",resourceNames="driver-toolkit",verbs=get;list;watch
```

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"fmt"
+	"sync"
+)
+
+//go:generate mockgen -source=syncronizedmap.go -package=syncronizedmap -destination=mock_syncronizedmap.go
+
+type KernelOsDtkMapping interface {
+	NewNodeInfo(kernelVersion, osImage string)
+	NewImageStreamInfo(osImage, dtkImage string)
```
qbarrand, line 12: I think the following names would be more explicit for a map:
```suggestion
	SetNodeInfo(kernelVersion, osImage string)
	SetImageStreamInfo(osImage, dtkImage string)
```
Or maybe
```suggestion
	SetKernelToOS(kernelVersion, osImage string)
	SetOSToDTKImage(osImage, dtkImage string)
```

internal/syncronizedmap/syncronizedmap.go
```diff
@@ -0,0 +1,67 @@
...
+
+	osImage, ok := skom.kernelToOs[kernelVersion]
+	if !ok {
+		return "", fmt.Errorf("could not find kernel %s in kernel --> OS mapping", kernelVersion)
+	}
+	dtk, ok := skom.osToDtk[osImage]
+	if !ok {
+		return "", fmt.Errorf("could not find osImage %s in OS --> DTK mapping", osImage)
+	}
+	return dtk, nil
```
qbarrand, line 66: Let us lock the OS --> DTK lock as late as possible:
```suggestion
	skom.kernelToOsMutex.RLock()
	defer skom.kernelToOsMutex.RUnlock()

	osImage, ok := skom.kernelToOs[kernelVersion]
	if !ok {
		return "", fmt.Errorf("could not find kernel %s in kernel --> OS mapping", kernelVersion)
	}
	
	skom.osToDtkMutext.RLock()
	defer skom.osToDtkMutext.RUnlock()

	dtk, ok := skom.osToDtk[osImage]
	if !ok {
		return "", fmt.Errorf("could not find osImage %s in OS --> DTK mapping", osImage)
	}
	return dtk, nil
```
Ideally we would also release the kernel --> OS one earlier than at the end of the function, but that is error-prone since we want to immediately return an error if the kernel key does not exist.

internal/filter/filter.go
```diff
@@ -46,12 +48,38 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
 	)
 }
 
+// This filter will queue the request in the following cases:
```
qbarrand, line 51: If we write a comment, let us follow the Go format:
```suggestion
// NodeKernelReconcilerPredicate will queue the request in the following cases:
```

internal/filter/filter.go
```diff
@@ -46,12 +48,38 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
...
+}
 
-	return predicate.And(skipDeletions, labelMismatch)
+func (f *Filter) ImageStreamReconcilerPredicate() predicate.Predicate {
+
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newTags := map[string]string{}
+			oldTags := map[string]string{}
+			for _, t := range e.ObjectNew.(*imagev1.ImageStream).Status.Tags {
```
qbarrand, line 74: Why the status tags and not the spec ones?

controllers/imagestream_reconciler.go
```diff
@@ -0,0 +1,58 @@
...
+		filter:             filter,
+		kernelOsDtkMapping: kernelOsDtkMapping,
+	}
+}
+
+func (r *ImageStreamReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	is := imagev1.ImageStream{}
+
+	if err := r.client.Get(ctx, types.NamespacedName{Name: req.Name}, &is); err != nil {
```
qbarrand, line 37:
```suggestion
	if err := r.client.Get(ctx, req.NamespacedName, &is); err != nil {
```

controllers/imagestream_reconciler.go
```diff
@@ -0,0 +1,58 @@
...
+		return ctrl.Result{}, fmt.Errorf("could not get imagestream %v: %v", req.Name, err)
+	}
+
+	for _, t := range is.Status.Tags {
```
qbarrand, line 41: Why the status tags and not the spec ones?

qbarrand (commented):
Final comments 🙂

controllers/imagestream_reconciler.go
ybettan: Fixed.

internal/syncronizedmap/syncronizedmap.go
ybettan: Prefer the first option. Fixed.

internal/syncronizedmap/syncronizedmap.go
ybettan: Don.e

controllers/imagestream_reconciler.go
ybettan: Done.

internal/filter/filter.go
ybettan: Done.

internal/filter/filter.go
ybettan: Fixed.

ybettan:
This PR was tested manually using this module:
```
apiVersion: kmm.sigs.k8s.io/v1beta1
kind: Module
metadata:
  name: kmm-ci-build
spec:
  moduleLoader:
    container:
      modprobe:
        moduleName: kmm_ci_a
      kernelMappings:
        - literal: 4.18.0-372.26.1.el8_6.x86_64
          containerImage: quay.io/ybettan/kernel-module-management:non-existing
          pull:
            insecure: true
          build:
            pull:
              insecure: true
            push:
              insecure: true
            dockerfile: |
              ARG DTK_AUTO
              FROM ${DTK_AUTO}
  selector:
    kubernetes.io/hostname: test-infra-cluster-7363ce8a-master-0

```

yevgeny-shnaidman:
@ybettan maybe you can push this test Module into some directory, like examples?

PR #107, opened 2022-11-17
 Garbage collection for successfull build jobs 

chr15p:
This PR contains the following changes:
    1) implement garbage collection for successful build jobs and pods
    2) refactoring jobhelper package and moving build code to use it
    3) refactoring/adding unit-test


main.go
```diff
@@ -23,6 +23,7 @@ import (
 	"os"
 	"runtime/debug"
 
+	//"github.com/rh-ecosystem-edge/kernel-module-management/internal/utils"
```
yevgeny-shnaidman, line 26: lets remove it if not needed

PR #109, opened 2022-11-18
Move to ubi-micro

qbarrand:
Move the manager binary to /usr/local/bin ([MGMT-12589](https://issues.redhat.com//browse/MGMT-12589)).
Use a TLS bundle provided by OpenShift ([MGMT-12594](https://issues.redhat.com//browse/MGMT-12594)).
Upgrade to the latest operator-sdk.

/cc @fabiendupont @cdvultur @mresvanis

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,10 +25,9 @@ COPY .git .git
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make manager
 
-FROM registry.ci.openshift.org/ocp/4.11:base
-WORKDIR /
-COPY --from=builder /workspace/manager .
-RUN useradd -r kmm
-USER kmm:kmm
+FROM registry.redhat.io/ubi8/ubi-micro:8.7
 
-ENTRYPOINT ["/manager"]
+COPY --from=builder /workspace/manager /usr/local/bin
+USER 65534:65534
+
+ENTRYPOINT ["/usr/local/bin/manager"]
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -25,10 +25,9 @@ COPY .git .git
...
-COPY --from=builder /workspace/manager .
-RUN useradd -r kmm
-USER kmm:kmm
+FROM registry.redhat.io/ubi8/ubi-micro:8.7
 
-ENTRYPOINT ["/manager"]
+COPY --from=builder /workspace/manager /usr/local/bin
+USER 65534:65534
+
+ENTRYPOINT ["manager"]
```
fabiendupont, line 33: I usually use an absolute path, just in case another program with the same name exists a higher priority path, like `/usr/bin`.

qbarrand: I used `manager` to remain compatible with possible future changes, but your point makes sense. PTAL.

bundle/manifests/kmm-operator.clusterserviceversion.yaml
```diff
@@ -239,8 +247,8 @@ spec:
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
                 command:
-                - /manager
-                image: quay.io/edge-infrastructure/kernel-module-management:latest
+                - /usr/local/bin/manager
+                image: quay.io/quba/kmmo:latest
```
cdvultur, line 251: Why do you point to your quay account ? Maybe is from testing ?

qbarrand: Should be fixed - apologies

mresvanis (approved):

PR #110, opened 2022-11-21
Updating the error in case a build fails.

ybettan:
In case a build fails, we should return the logs snipped of the build pod instead of nil.

### DIFF
--- a/internal/build/buildconfig/manager.go
+++ b/internal/build/buildconfig/manager.go
@@ -85,7 +85,7 @@ func (bcm *buildConfigManager) Sync(ctx context.Context, mod kmmv1beta1.Module, 
 	case buildv1.BuildPhaseNew, buildv1.BuildPhasePending, buildv1.BuildPhaseRunning:
 		return build.Result{Status: build.StatusInProgress, Requeue: true}, nil
 	case buildv1.BuildPhaseFailed:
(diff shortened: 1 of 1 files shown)


ybettan:
Here is how the new error looks like:
```
E1121 13:35:13.666946       1 controller.go:326]  "msg"="Reconciler error" "error"="failed to handle build for kernel version 4.18.0-372.26.1.el8_6.x86_64: could not synchronize the build: buildConfig failed: Copying blob sha256:a6f4193590ca52a615e071ee13b1193b00846684d96a07734133ba401c0b5bc6\nCopying blob sha256:a69f0d32ba160f015d04db3758a223e1bae9e150f9e646fe293d38b1e69be491\nCopying blob sha256:725a55c4212630f1b818ee1e82c5f7a9e4ed42456f19ea13b052da427cfdff82\nWarning: Push failed, retrying in 5s ...\nerror: build error: Failed to push image: writing blob: in...horized: access to the requested resource is not authorized" "controller"="module" "controllerGroup"="kmm.sigs.k8s.io" "controllerKind"="Module" "module"={"name":"simple-kmod-bld-quay","namespace":"default"} "name"="simple-kmod-bld-quay" "namespace"="default" "reconcileID"="0f80950b-58e4-486c-b6fd-90ac69c69d9d"
```

PR #122, opened 2022-11-24
Run GitStream in its container image

qbarrand:
Run GitStream every 10 minutes and each time a PR or an issue is closed.


iranzo (commented):
lgtm

PR #136, opened 2022-11-25
Signing jobmanager requires permissions to add/delete/list jobs

chr15p:
Signing jobmanager needs to be able to create the signing job and tell when it has completed.

(These permissions are in upstream for the kaniko build job, but weren't needed here for buildconfig. Now we need them here as well)

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -108,6 +108,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterroles,verbs=bind,resourceNames=module-loader;device-plugin
 //+kubebuilder:rbac:groups="build.openshift.io",resources=builds,verbs=get;list;create;delete;watch;patch
+//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
 
 // Reconcile lists all nodes and looks for kernels that match its mappings.
 // For each mapping that matches at least one node in the cluster, it creates a DaemonSet running the container image
--- a/bundle/manifests/kmm-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kmm-operator.clusterserviceversion.yaml
@@ -51,23 +51,24 @@ spec:
           - patch
           - watch
         - apiGroups:
-          - build.openshift.io
+          - batch
           resources:
-          - buildconfigs
+          - jobs
           verbs:
           - create
           - delete
-          - get
           - list
-          - patch
           - watch
         - apiGroups:
           - build.openshift.io
           resources:
           - builds
           verbs:
+          - create
+          - delete
           - get
           - list
+          - patch
           - watch
         - apiGroups:
           - ""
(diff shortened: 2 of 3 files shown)


qbarrand:
I think we should rather add a `//kubebuilder:rbac` comment like we do [upstream](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/controllers/module_reconciler.go#L100).
`make manifests bundle` will then add those new RBAC rules to the right files.

controllers/module_reconciler.go
```diff
@@ -108,6 +108,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterroles,verbs=bind,resourceNames=module-loader;device-plugin
 //+kubebuilder:rbac:groups="build.openshift.io",resources=builds,verbs=get;list;create;delete;watch;patch
+//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
```
qbarrand, line 111: Please run `make manifests bundle` and add the generated files to the PR.

ybettan: FYI @enriquebelarte.

Seems like `make bundle` is also needed for making sure we are pushing generated files.

ybettan:
> Now we need them here as well

Can you please add what you wrote in the PR description in the commit message as well?

Because we moved to `builds` instead of `buildconfigs`? If this is the case can we please add it to the commit message as well?

controllers/module_reconciler.go
qbarrand: `make manifests bundle` runs both targets 🙂

qbarrand:
> Because we moved to builds instead of buildconfigs?

Nope, because upstream already uses jobs to build and hence has the required permissions.

controllers/module_reconciler.go
ybettan: Yes. @enriquebelarte and I were talking about a job to make sure generated files aren't forgotten from PRs and we only talked about `make generate manifests` so I wan't to mention that we also need `make bundle` in the job.

This was intended for Enrique :)

enriquebelarte: OK, I'll create a new PR for that.

chr15p: ok I *think* I have this right, turns out if you rebase onto main first running `make manifests bundle` generates a load of other changes that are not related to this PR, so I have un-rebased and (hopefully!) I just have my changes, but I think that means we need another PR that generates the other changes (related to moving form dockerfiles to config maps iirc)

chr15p:
> > Now we need them here as well
> 
> Can you please add what you wrote in the PR description in the commit message as well? Having fully contained commit messages when using `Git blame` helps a lot.

Done. One day I'll get a commit message/PR message you like! 
(to be clear, you are right, getting a clear message is important and one day I'll get the hang of writing them :-)  )

controllers/module_reconciler.go
ybettan: Indeed. We already have a job u/s to make sure generated files are not forgotten. Once it is pulled m/s by the new sync tool we will have that validation as well m/s.

ybettan:
@chr15p TY for updating the commit message :) 
/lgtm

PR #145, opened 2022-11-25
Install Go manually in the GitStream container

qbarrand:
The setup-go action does not work on Alpine Linux, the base image for the GitStream container.
Install Go manually and leverage caching to avoid downloading at every run.

### DIFF
--- a/.github/workflows/gitstream.yml
+++ b/.github/workflows/gitstream.yml
@@ -16,6 +16,8 @@ jobs:
   gitstream:
     name: Run GitStream
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


ybettan:
> The setup-go action does not work on Alpine Linux, the base image for the GitStream container.

Why do we care about Alpine? Aren't github runner running Ubuntu?

qbarrand:
@ybettan We are running the container in the action to avoid building the image on every run: https://github.com/rh-ecosystem-edge/kernel-module-management/blob/main/.github/workflows/gitstream.yml#L20-L21

PR #147, opened 2022-11-27
Moving Build Dockerfile to be defined in the ConfigMap

yevgeny-shnaidman:
Currently dockerfile is defined a multiline string in the Build field of the KernelMapping. This PR does the following: 1) dockerfile is now defined in a dedicated configmap (created by customer)
   with the key being: dockerfile
2) Build struct now contains reference to the ConfigMap, instead of
   a string


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/a3d7e28640a598065b9e9ec303c95ce66b9088be

controllers/module_reconciler.go
```diff
@@ -103,6 +103,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list
```
ybettan, line 106: I though we also need to watch in case the user update the configMap before the image finished to build and was pushed.

yevgeny-shnaidman: it is a corner case. We will get the ConfigMap update once the Build fails

PR #150, opened 2022-11-28
Remove /usr/lib references at DaemonSet (#50)

qbarrand:
In most cases shared libraries directory /lib is a dynamic link to
 /usr/lib but in some other cases as Ubuntu 18.04 LTS or Debian 11,
those directories have completely different content. In any case in-tree modules can be found on /lib/modules/KERNEL_VERSION/ so in order to avoid compatibility issues depending which OS Kubernetes nodes are running we should stick to /lib as a mount point for DaemonSet. Also nodeLibModulesPath has been restricted to /lib/modules/KERNEL_VERSION/.

/cc @enriquebelarte

Upstream-Commit: f7499feba6f2978d897b9b14c93a208b612131eb


qbarrand:
Fixes #148

PR #65, opened 2022-10-04
Add first doc and examples to labs

enriquebelarte:
Labs consisting on a document including different examples and

 use cases on how to build and/or load a kernel module and a device plugin.

PR #164, opened 2022-11-28
adding build verification to preflight (#65)

yevgeny-shnaidman:
In case image of the Module is not available or
is does not contain the correct driver version,
preflight will check if Build configuration exists in Module yaml, and if it does, then it will try to build it. In case build is successful, then the module is considered to be verified

Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/54ec19e957fa30e6faf425fb120aadf2e640251e

### DIFF
--- a/controllers/preflightvalidation_reconciler.go
+++ b/controllers/preflightvalidation_reconciler.go
@@ -120,7 +120,7 @@ func (r *PreflightValidationReconciler) runPreflightValidation(ctx context.Conte
 	for _, module := range modulesToCheck {
 		log.Info("start module preflight validation", "name", module.Name)
 
-		verified, message := r.preflight.PreflightUpgradeCheck(ctx, &module, pv.Spec.KernelVersion)
+		verified, message := r.preflight.PreflightUpgradeCheck(ctx, pv, &module, pv.Spec.KernelVersion)
 
 		log.Info("module preflight validation result", "name", module.Name, "verified", verified)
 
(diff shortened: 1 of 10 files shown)


config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/edge-infrastructure/kernel-module-management
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
```
ybettan, line 16: Forgotten.

controllers/preflightvalidation_reconciler.go
```diff
@@ -120,7 +120,7 @@ func (r *PreflightValidationReconciler) runPreflightValidation(ctx context.Conte
 	for _, module := range modulesToCheck {
 		log.Info("start module preflight validation", "name", module.Name)
 
-		verified, message := r.preflight.PreflightUpgradeCheck(ctx, &module, pv.Spec.KernelVersion)
+		verified, message := r.preflight.PreflightUpgradeCheck(ctx, pv, &module, pv.Spec.KernelVersion)
```
ybettan, line 123: Why do we need both `pv` and `pv.Spec.KernelVersion`?

ybettan:
@yevgeny-shnaidman Can you please elaborate on the differences vs u/s? The u/s commit seems very different from this one.

yevgeny-shnaidman:
> @yevgeny-shnaidman Can you please elaborate on the differences vs u/s? The u/s commit seems very different from this one.

@ybettan this is mainly due to differences in the Builds between upstream and Downstream. The Sync API is different...

controllers/preflightvalidation_reconciler.go
yevgeny-shnaidman: this is the way it was added to the upstream. We can probably remove it, but it is better to do it in synchronized PRs both to upstream and downastream

config/manager/kustomization.yaml
yevgeny-shnaidman: fixed

controllers/preflightvalidation_reconciler.go
ybettan: `pv.Spec.KernelVersion` isn't present u/s.
See https://github.com/kubernetes-sigs/kernel-module-management/blob/2642a7db8531314c6ae4f2504cd1e4c0dc69d3aa/controllers/preflightvalidation_reconciler.go#L123

PR #172, opened 2022-11-28
Adding support for image push during preflight build verification (#76)

yevgeny-shnaidman:
1) add a field to the Preflight CRD spec , that signifies if customer
   wants to push the image during preflight verification
2) add support for image push during preflight Build stage 3) add Generator filter to preflight Watch, to skip being scheduled
   on Preflight status changes


yevgeny-shnaidman:
Upstream PR for cherry-pick:
https://github.com/kubernetes-sigs/kernel-module-management/commit/9b2db6ca2a92548722887f41f245f4860f4a1b8f

PR #212, opened 2022-12-01
Do not pass the kernel-OS-DTK mapping around

qbarrand:
Make the kernel version --> OS version --> DTK image an instance variable of `build.Maker` instead of passing it around.

/cc @mresvanis @ybettan @yevgeny-shnaidman

### DIFF
--- a/internal/preflight/preflight.go
+++ b/internal/preflight/preflight.go
@@ -10,7 +10,6 @@ import (
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/module"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/registry"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/statusupdater"
-	"github.com/rh-ecosystem-edge/kernel-module-management/internal/syncronizedmap"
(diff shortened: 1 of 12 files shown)


internal/preflight/preflight.go
```diff
@@ -90,21 +90,19 @@ type preflightHelperAPI interface {
...
-	authFactory        auth.RegistryAuthGetterFactory
-	kernelOsDtkMapping syncronizedmap.KernelOsDtkMapping
+	client      client.Client
+	registryAPI registry.Registry
+	buildAPI    build.Manager
+	authFactory auth.RegistryAuthGetterFactory
 }
 
 func newPreflightHelper(client client.Client, buildAPI build.Manager, registryAPI registry.Registry,
 	authFactory auth.RegistryAuthGetterFactory, kernelOsDtkMapping syncronizedmap.KernelOsDtkMapping) preflightHelperAPI {
```
yevgeny-shnaidman, line 100: it needs to be removed from function parameters

qbarrand: Done, PTAL.

PR #215, opened 2022-12-02
refactoring GetHeaderDataFromLayer function

yevgeny-shnaidman:
GetHeaderDataFromLayer function is also used to verify that a file(header) exists in an image(layer). it check for header existence and reads its input into a buffer. Since this function can be used to verify a kernel module file existence, it can be problematic to read a whole file into a memory, since kernel module file can be a big one. This PR split this function into an inner implementation that checks header existence and returns a io.Reader to it, and the upper implementation that is actually using the io.Reader to read the data

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -2,7 +2,6 @@ package registry
 
 import (
 	"archive/tar"
-	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
@@ -113,8 +112,15 @@ func (r *registry) LastLayer(ctx context.Context, image string, po *kmmv1beta1.P
 func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool {
 	// in layers headers, there is no root prefix
 	fullPath := filepath.Join(strings.TrimPrefix(pathPrefix, "/"), modulesLocationPath, kernelVersion, moduleFileName)
-	_, err := r.GetHeaderDataFromLayer(layer, fullPath)
-	return err == nil
+
+	// if getHeaderReaderFromLayer does not return an error, it means that the file exists in the layer,
+	// and that's all the indication that we need
+	_, readerCloser, err := r.getHeaderReaderFromLayer(layer, fullPath)
+	if err != nil {
+		return false
+	}
+	readerCloser.Close()
+	return true
 }
 
 func (r *registry) getPullOptions(ctx context.Context, image string, po *kmmv1beta1.PullOptions, registryAuthGetter auth.RegistryAuthGetter) (*RepoPullConfig, error) {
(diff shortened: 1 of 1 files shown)


internal/registry/registry.go
```diff
@@ -224,22 +227,31 @@ func (r *registry) getLayersDigestsFromManifestStream(manifestStream []byte) ([]
...
+	defer readerCloser.Close()
+	buff, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read header %s data: %v", headerName, err)
+	}
+	return buff, nil
+}
 
-	gr, err := gzip.NewReader(targz)
+func (r *registry) getHeaderReaderFromLayer(layer v1.Layer, headerName string) (io.Reader, io.ReadCloser, error) {
```
qbarrand, line 242:
```suggestion
func (r *registry) getHeaderReaderFromLayer(layer v1.Layer, headerName string) (io.Reader, io.Closer, error) {
```
I guess we don't want callers to be able to keep reading the layer.

qbarrand: I think providing this method with a callback function is safer solution. That way we can close everything we need ourselves.

yevgeny-shnaidman: This is an inner function of the registry package. User does not need to know what and how we are doing inside. Currently there are 2 methods that are using it: one is GetHeaderDataFromLayer, which calls this method, and then uses the io.Reader to read the data, closes  the io.ReadCloser and return the data to user. The second one is VerifyModuleExists, which calls this method, and then just closes the io.ReadCloser, since it only wants to verify that the header exists. I think that providing a callback will only complicate the flow

internal/registry/registry.go
```diff
@@ -113,8 +112,12 @@ func (r *registry) LastLayer(ctx context.Context, image string, po *kmmv1beta1.P
...
 	// in layers headers, there is no root prefix
 	fullPath := filepath.Join(strings.TrimPrefix(pathPrefix, "/"), modulesLocationPath, kernelVersion, moduleFileName)
-	_, err := r.GetHeaderDataFromLayer(layer, fullPath)
-	return err == nil
+	_, readerCloser, err := r.getHeaderReaderFromLayer(layer, fullPath)
+	if err != nil {
+		return false
+	}
+	readerCloser.Close()
+	return true
```
ybettan, line 120:
```suggestion
	_, readerCloser, err := r.getHeaderReaderFromLayer(layer, fullPath)
	defer readerCloser.Close()
        return err == nil
```

internal/registry/registry.go
```diff
@@ -113,8 +112,12 @@ func (r *registry) LastLayer(ctx context.Context, image string, po *kmmv1beta1.P
 func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool {
 	// in layers headers, there is no root prefix
 	fullPath := filepath.Join(strings.TrimPrefix(pathPrefix, "/"), modulesLocationPath, kernelVersion, moduleFileName)
-	_, err := r.GetHeaderDataFromLayer(layer, fullPath)
-	return err == nil
+	_, readerCloser, err := r.getHeaderReaderFromLayer(layer, fullPath)
```
ybettan, line 115: Can we add a comment saying that if we didn't get any error it means that the file exists? (we are not using the reader so it is less explicit than before).

internal/registry/registry.go
```diff
@@ -224,22 +227,31 @@ func (r *registry) getLayersDigestsFromManifestStream(manifestStream []byte) ([]
...
+	}
+	return buff, nil
+}
 
-	gr, err := gzip.NewReader(targz)
+func (r *registry) getHeaderReaderFromLayer(layer v1.Layer, headerName string) (io.Reader, io.ReadCloser, error) {
+	layerreader, err := layer.Uncompressed()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create reader from targz: %w", err)
+		return nil, nil, fmt.Errorf("failed to get layerreader from layer: %w", err)
```
ybettan, line 245:
```suggestion
		return nil, nil, fmt.Errorf("failed to get layerreader from layer: %v", err)
```

internal/registry/registry.go
```diff
@@ -224,22 +227,31 @@ func (r *registry) getLayersDigestsFromManifestStream(manifestStream []byte) ([]
...
 	}
 	// err ignored because we're only reading
-	defer gr.Close()
+	defer func() {
+		if err != nil {
```
ybettan, line 249: Seems like this func is enclosing `err`. It may change later in this function to contain the value of another action error instead of the intended one.

internal/registry/registry.go
yevgeny-shnaidman: I cannot use your suggestion since  in case getHeaderReaderFromLayer returns an error, readerCloser will be nil.

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry.go
yevgeny-shnaidman: fixed

internal/registry/registry.go
yevgeny-shnaidman: As discussed, this is the intention: we want to close ReadCloser in case of any error in the function

PR #229, opened 2022-12-07
Only reconcile openshift/driver-toolkit

qbarrand:
Only watch the openshift/driver-toolkit ImageStream.
Filter out reconciliation requests for other ImageStreams.
Use a namespace-scoped client.

/cc @ybettan @enriquebelarte @yevgeny-shnaidman

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"runtime/debug"
 
-	//"github.com/rh-ecosystem-edge/kernel-module-management/internal/utils"
 	buildv1 "github.com/openshift/api/build/v1"
 	imagev1 "github.com/openshift/api/image/v1"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/auth"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -218,7 +219,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = controllers.NewImageStreamReconciler(client, filter, kernelOsDtkMapping).SetupWithManager(mgr); err != nil {
+	const dtkImageStreamNamespace = "openshift"
```
yevgeny-shnaidman, line 222: WDYT about moving namespace and name into constants.go, and make ImageReconciler create the nsn?

qbarrand: I think that would only make sense if we renamed the reconciler into something like `DTKImageStreamReconciler`. WDYT?

yevgeny-shnaidman: since this is what it does anyway ( i don't think that it's code can work in case it is passed a different image stream), then let's rename it in a different PR

PR #262, opened 2022-12-09
Adding PreflightOCP controller

yevgeny-shnaidman:
Preflight OCP controller is responsible for initiating the preflight flow in OCP cluster. The difference between PreflightOCP and Preflight is that the input to PreflightOCP is an OCP release image and not kernel version. In order to utilize the already existing Preflight controller, the flow of the PreflightOCP controller is as following:
1) extract DTK, OS version and kernel version from the input OCP image 2) update the kernel dtk mapping with the extracted data 3) create a new Preflight CR with the extracted kernel version 4) monitor status updates of the Preflight and update Preflight OCP
   Status with the data taken from Preflight CR Status

This way we are not interfering with the upstream flow of the Preflight, and the PreflightOCP controller is just an upper layer that prepares data for the Preflight flow, and then just copies its statuses


controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,271 @@
...
+func (r *PreflightValidationOCPReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("preflightvalidationocp").
+		For(&kmmv1beta1.PreflightValidationOCP{}, builder.WithPredicates(filter.PreflightOCPReconcilerUpdatePredicate())).
+		Watches(
+			&source.Kind{Type: &kmmv1beta1.PreflightValidation{}},
+			handler.EnqueueRequestsFromMapFunc(r.filter.EnqueueAllPreflightValidationsOCP),
+		).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
```
mresvanis, line 97: This is the default value for the manager's respective global [option](https://github.com/kubernetes-sigs/controller-runtime/blob/92611cbdcaa9eb58f3e05697b20c0b282cb6447c/pkg/controller/controller.go#L39), but also the controller's default [option](https://github.com/kubernetes-sigs/controller-runtime/blob/92611cbdcaa9eb58f3e05697b20c0b282cb6447c/pkg/internal/controller/controller.go#L47). 

Do you emphasize here the fact that this reconciler should strictly stick with just one workqueue event processed at any time  (with its current implementation)? 

I have absolutely no problem with having this stated explicitly here, but I think increasing concurrency in any of the reconcilers would require very good verification that their implementation is "thread"-safe. So maybe we could skip it. Whatever you think is best. 

(mostly wrote this comment to make sure I understand all the aspects correctly)

controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,271 @@
...
+		}).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=preflightvalidations,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=preflightvalidations/status,verbs=get;list;watch
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=preflightvalidations/finalizers,verbs=update
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=preflightvalidationsocp,verbs=get;list;watch;create;update;patch;delete
```
mresvanis, line 105: If I'm not mistaken we could remove `create` and `delete` permissions on `PreflightValidationsOCP` CRs, as only the users would perform such operations.  WDYT?

controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,271 @@
...
+			Name:      pvo.Name,
+			Namespace: pvo.Namespace,
+		},
+		Spec: kmmv1beta1.PreflightValidationSpec{
+			KernelVersion:  kernelVersion,
+			PushBuiltImage: pvo.Spec.PushBuiltImage,
+		},
+	}
+
+	err = controllerutil.SetControllerReference(pvo, &pv, r.scheme)
```
mresvanis, line 204:
```suggestion
	if err := controllerutil.SetControllerReference(pvo, &pv, r.scheme); err != nil {
```

controllers/preflightvalidationocp_reconciler.go
qbarrand: > I have absolutely no problem with having this stated explicitly here

I do 🙂 I think it might bring some confusion to someone reading this. "Why would they explicitly specify it? Are the defaults overridden anywhere?"

controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,271 @@
...
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, "preflight validation ocp reconcile failed to find object")
+		return ctrl.Result{}, err
+	}
+
+	err = r.runPreflightValidationOCP(ctx, &pvo)
+	if err != nil {
+		log.Error(err, "runPreflightValidationOCP failed")
+		return ctrl.Result{}, err
```
qbarrand, line 128:
```suggestion
		return ctrl.Result{}, fmt.Errorf("runPreflightValidationOCP failed: %v", err)
```

controllers/preflightvalidationocp_reconciler_test.go
```diff
@@ -0,0 +1,231 @@
...
+					Namespace: "pvo namespace",
+				},
+				Spec: kmmv1beta1.PreflightValidationOCPSpec{
+					ReleaseImage:   "ocpReleaseImage",
+					PushBuiltImage: true,
+				},
+			}
+		})
+
+		It("good flow", func() {
```
mresvanis, line 208: Would we like to have also a couple of the errored flows of this method tested? For example, we could test an error on [this](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/262/files#diff-896d99d61b4773ad699ff552b435c9e1a4622f44edd4bc1a87d1e7151f527596R172-R175) and [this](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/262/files#diff-896d99d61b4773ad699ff552b435c9e1a4622f44edd4bc1a87d1e7151f527596R180-R182). WDYT?

PS. I realize that it's not straight forward, since we are mocking packages that are not directly called by `preparePreflightValidation()`, but I think it would make sense to have the errored flows tested as well.

controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,271 @@
...
+	pv := &kmmv1beta1.PreflightValidation{}
+	err := r.client.Get(ctx, nsn, pv)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Compatible PreflightValidation not found, creating")
+			pv, err = r.preparePreflightValidation(ctx, pvo)
+			if err != nil {
+				return fmt.Errorf("failed to prepare the data for compatible PreflightValidation: %v", err)
+			}
+			err = r.client.Create(ctx, pv)
```
qbarrand, line 148: Isn't [`CreateOrPatch`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch) and option here? We use it extensively in other controllers.

controllers/preflightvalidationocp_reconciler.go
qbarrand: I also prefer that style.

qbarrand: I understand that several `PreflightValidationOCP` objects can own a `PreflightValidation`? If so, I think we should use [`SetOwnerReference`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetOwnerReference) instead.

internal/filter/filter.go
```diff
@@ -228,6 +228,19 @@ func (f *Filter) EnqueueAllPreflightValidations(mod client.Object) []reconcile.R
...
+	pvo := kmmv1beta1.PreflightValidationOCP{}
+
+	nsn := types.NamespacedName{Name: pv.GetName(), Namespace: pv.GetNamespace()}
+	err := f.client.Get(context.Background(), nsn, &pvo)
+	if err != nil {
+		logger.Error(err, "failed to find PVO related to", "preflight", nsn)
+		return []reconcile.Request{}
+	}
+	return []reconcile.Request{reconcile.Request{NamespacedName: nsn}}
+}
```
qbarrand, line 242: Is this necessary if PVOs are listed as owners of the PV? We could then just use `Owns(&kmmv1beta1.PreflightValidation{})`.

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: I always like to state it explicitly, since not everybody remembers the default setting. It always helps to filter out debug versions that say: "maybe we have a concurrency issue here"

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: I prefer to set the err != nil on a different line, i think it is a little bit more readable

controllers/preflightvalidationocp_reconciler_test.go
yevgeny-shnaidman: We have an extensive error flow for both getKernelVersionAndOSFromDTK and getDTKFromImage ( since most of the extraction and substition logic is there.  preparePreflightValidation is pretty straightforward, so i think that only verifying good flow should be sufficient ( it is done mainly to test that we are calling getKernelVersionAndOSFromDTK, getDTKFromImage and kernelOsDtkMapping in correct order

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: fixed

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: Currently i do not see a scenario where user would like to update a Preflight or PreflightOCP. If user wants another kernel version/release image verifed, then he should create a new Preflight/ Preflight OCP CR

controllers/preflightvalidationocp_reconciler_test.go
mresvanis: I understand your point and it is minor and straight forward, but it would make it easier for another developer to contribute to `preparePreflightValidation()` if needed.

> Just as a minor comment, anyhow it is uncommon to test private methods/functions, but I do get why we're doing it here.

controllers/preflightvalidationocp_reconciler.go
mresvanis: As I mentioned above, I have no issue with it being explicit here, if we decide to leave it as is. IMHO:

1. having `MaxConcurrentReconciles=1` is the strict case for all KMM reconcilers currently
2. it's the other way round regarding concurrency and reconcilers, whoever tries to go for more than `MaxConcurrentReconciles > 1` would need to explicitly justify, verify every bit of code and unit-test extensively

internal/filter/filter.go
yevgeny-shnaidman: yes, it is necessary, when deleting the PreflightOCP, the Preflight is also deleted by kubernetes. Changed the For to Owns

yevgeny-shnaidman: One PreflightOCP owns exactly one Preflight

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: fixed

controllers/preflightvalidationocp_reconciler.go
yevgeny-shnaidman: @qbarrand one PreflightValidationOCP owns exactly one Preflight

qbarrand:
@yevgeny-shnaidman please add an entry for the new CRD in the `PROJECT` file.

controllers/preflightvalidationocp_reconciler.go
qbarrand: That sounds a bit counter-intuitive to me, unless we make the field immutable (which seems only doable with a webhook). If you insist, we will need to make sure that this ends up in the docs.

yevgeny-shnaidman:
> @yevgeny-shnaidman please add an entry for the new CRD in the `PROJECT` file.

Fixed

qbarrand:
/lgtm
/hold
@mresvanis if you are happy as well, please `/unhold`.

PR #265, opened 2022-12-12
Mount the OpenShift service certificate

qbarrand:
Create a new ConfigMap and annotate it with `service.beta.openshift.io/inject-cabundle=true`.
Move all OCP manifest customizations to a separate patch.

Fixes [MGMT-12141](https://issues.redhat.com/browse/MGMT-12141)

/cc @mresvanis @yevgeny-shnaidman @ybettan


yevgeny-shnaidman:
@qbarrand i am missing something: how it is related to [MGMT-12141]?

PR #267, opened 2022-12-13
Add e2e tests

enriquebelarte:
Add e2e tests to prow. e2e-incluster-build.

### DIFF
--- a/ci/module-kmm-ci-build.template.yaml
+++ b/ci/module-kmm-ci-build.template.yaml
@@ -10,22 +10,19 @@ spec:
         moduleName: kmm_ci_a
       kernelMappings:
         - literal: KVER_CHANGEME
-          containerImage: registry.minikube/kmm-kmod:local
+          containerImage: image-registry.openshift-image-registry.svc:5000/default/kmm-kmod:local
           registryTLS:
             insecure: true
           build:
             buildArgs:
               - name: CI_BUILD_ARG
                 value: some-build-arg
-            baseImageRegistryTLS:
-              insecure: true
             secrets:
               - name: build-secret
             dockerfileConfigMap:
               name: kmm-kmod-dockerfile
-            
             # Optional. If kanikoParams.tag is empty, the default value will be: 'latest'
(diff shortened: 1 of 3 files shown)


ci/module-kmm-ci-build.template.yaml
```diff
@@ -10,7 +10,7 @@ spec:
         moduleName: kmm_ci_a
       kernelMappings:
         - literal: KVER_CHANGEME
-          containerImage: registry.minikube/kmm-kmod:local
+          containerImage: registry.ci.openshift.org/rh-ecosystem-edge/kmm-kmod:local
```
qbarrand, line 13: I think we rather want to push the image to the internal OpenShift Registry?
```suggestion
          containerImage: image-registry.openshift-image-registry.svc:5000/default/kmm-kmod:local
```

ci/module-kmm-ci-build.template.yaml
```diff
@@ -23,9 +23,8 @@ spec:
               - name: build-secret
             dockerfileConfigMap:
               name: kmm-kmod-dockerfile
-            
             # Optional. If kanikoParams.tag is empty, the default value will be: 'latest'
             kanikoParams:
               tag: "debug"
   selector:
-    kubernetes.io/hostname: minikube
+    kubernetes.io/hostname: kmm-ci
```
qbarrand, line 30: Do we have such a node on the test cluster?

enriquebelarte: I pretended to label the node before applying:

https://github.com/rh-ecosystem-edge/kernel-module-management/pull/267/files#diff-d954caf0668f1d5539f11597234490aa6aadf4e53c4e175ba384b11244dbd21eR12

qbarrand: I think we should use another label key then. `kubernetes.io/hostname` is managed by Kubernetes and might be overwritten.

qbarrand (approved):

qbarrand (approved):

PR #295, opened 2022-12-15
Check e2e check kmm function

enriquebelarte:


ci/prow/e2e-crd-variable
```diff
@@ -28,10 +28,10 @@ echo " Add an kmm-ci Module that contains a valid mapping..."
 oc apply -f ci/module-kmm-ci-variable.yaml
 
 echo "Check that the module gets loaded on the node..."
-timeout 1m bash -c 'until check_kmm_kmod; do sleep 3; done'
+timeout 1m bash -c 'until check_kmm_mod; do sleep 3; done'
```
ybettan, line 31: I would change the function name instead `check_kmm_kmod`.
Also we can fix the indention in the function in this PR.

ci/prow/e2e-incluster-build
```diff
@@ -31,7 +31,7 @@ echo "Wait for the job to be created..."
 timeout 1m bash -c 'until oc get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
 
 echo "Check that the module gets loaded on the node..."
-timeout 1m bash -c 'until check_kmm_kmod; do sleep 3; done'
+timeout 1m bash -c 'until check_kmm_mod; do sleep 3; done'
```
ybettan, line 34: Can we extract `check_kmm_kmod` function to another file and use it? We are just duplicating it ATM.

enriquebelarte: why? it is a one-line function and extract it to another file would imply to source it from the scripts, losing clarity IMHO

ci/prow/e2e-crd-variable
```diff
@@ -2,8 +2,8 @@
 
 set -euxo pipefail
 
-check_kmm_mod () {
-oc debug node/$NODE -T -- chroot /host sh -c "lsmod | grep kmm"
+check_kmm_kmod () {
+  oc debug node/$NODE -T -- chroot /host sh -c "lsmod | grep kmm"
```
ybettan, line 6: Can we extract it to `check-kmm-kmod` script and source it everywhere instead of duplicating it?

ci/prow/e2e-crd-variable
```diff
@@ -2,13 +2,12 @@
...
 set -euxo pipefail
 
-check_kmm_mod () {
-oc debug node/$NODE -T -- chroot /host sh -c "lsmod | grep kmm"
-}
-
 echo "Get first node..."
 NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
 
+echo "Load check_kmm_kmod function..."
```
ybettan, line 8: Can we remove that comment? `source` is self documenting.

ci/prow/e2e-crd-variable
```diff
@@ -2,13 +2,12 @@
...
+source check_kmm_kmod
```
ybettan, line 9: This is only "combining the new files", you still need to use the function somewhere.
Also can we source `check-kmm-kmod` at the top of all files?

ybettan: Same comment for all files.

enriquebelarte: Function name did not change so it is still called.

PR #300, opened 2022-12-18
Fixes call to check_kmm_mod function

enriquebelarte:
Function check\_kmm\_kmod is exported in order to be used by e2e test scripts. This function checks in cluster nodes if kmm-kmod module is loaded or not.

### DIFF
--- a/ci/prow/e2e-crd-variable
+++ b/ci/prow/e2e-crd-variable
@@ -2,10 +2,11 @@
 
 set -euxo pipefail
 
-echo "Get first node..."
-NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
-
 source ci/prow/check-kmm-kmod
+export -f check_kmm_kmod
+
+echo "Get first node..."
+export NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
 
 echo "Label a node to match selector in Module afterwards..."
 oc label node $NODE task=kmm-ci
--- a/ci/prow/e2e-prebuilt-kernel-module
+++ b/ci/prow/e2e-prebuilt-kernel-module
@@ -2,10 +2,11 @@
 
 set -euxo pipefail
 
-echo "Get first node..."
-NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
-
 source ci/prow/check-kmm-kmod
+export -f check_kmm_kmod
+
+echo "Get first node..."
+export NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
 
 echo "Label a node to match selector in Module afterwards..."
 oc label node $NODE task=kmm-ci
--- a/ci/prow/e2e-incluster-build
+++ b/ci/prow/e2e-incluster-build
@@ -2,16 +2,17 @@
 
 set -euxo pipefail
 
-echo "Deploy KMMO..."
-make deploy
+source ci/prow/check-kmm-kmod
+export -f check_kmm_kmod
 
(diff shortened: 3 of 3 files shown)


ci/prow/e2e-crd-variable
```diff
@@ -3,11 +3,10 @@
 set -euxo pipefail
 
 echo "Get first node..."
-NODE=$(oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'})
-
-source ci/prow/check-kmm-kmod
+oc get nodes -o jsonpath='{.items..metadata.name}' | awk {'print $1'} > nodename
```
ybettan, line 6: I think that if you `export NODE=...` instead of `NODE=...` it will work.

ci/prow/e2e-crd-variable
```diff
@@ -17,7 +16,7 @@ echo "Wait until the KMMO Deployment is Available..."
 oc wait --for condition=Available --timeout 1m deployments.apps -n kmm-operator-system kmm-operator-controller-manager
 
 echo "Check that the kmm_ci_a module is not loaded on the node..."
-if check_kmm_kmod; then
+if . ./ci/prow/check-kmm-kmod; then
```
ybettan, line 19: What does that command do?

ybettan:
Can you please describe in the commit message (and PR description) why this change is needed?

ci/prow/e2e-crd-variable
enriquebelarte: source is a synonym for dot/period '.' in bash. So it's just like source command

ybettan: Why don't we use `source` in that case? It is more explicit.

ci/prow/e2e-crd-variable
enriquebelarte: You are right. I'll remove the output to the temp file.

ci/prow/e2e-crd-variable
enriquebelarte: shorter?

ybettan: I see. Thanks for the explanation. The way I am seeing it (which is usually what we do in `go` as well):
* `ci/prow/check-kmm-kmod` should only contain the function definition (not run it)
* each test using that function should `source` the file at the begining of the script (same as we do for importing packages in `go`, `python`, etc...
* each test that requires `check_kmm_kmod` function should call it explicitely.

WDYT?

enriquebelarte: > I see. Thanks for the explanation. The way I am seeing it (which is usually what we do in `go` as well):
> 
> * `ci/prow/check-kmm-kmod` should only contain the function definition (not run it)
> * each test using that function should `source` (explicitely) the file at the begining of the script (same as we do for importing packages in `go`, `python`, etc...
> * each test that requires `check_kmm_kmod` function should call it explicitely.
> 
> WDYT?

We sourced the function at the beginning of each script before but as I explain in commit, source is lost when you call the function from a new session (bash -c), hence the change.

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -46,7 +47,7 @@ echo "Remove the Module..."
 oc delete -f module-kmm-ci.yaml
 
 echo "Check that the module gets unloaded from the node..."
-timeout 1m bash -c 'until ! check_kmm_kmod; do sleep 3; done'
+timeout 1m bash -c 'until ! check_kmm_mod; do sleep 3; done'
```
ybettan, line 50: Is this a typo?

enriquebelarte: Totally. Fixed, thank you.

PR #302, opened 2022-12-18
Fixing FIRWARE files/directories loading

yevgeny-shnaidman:
Currently the firmware path defined in the Module CR is copied into the /var/lib/firmware host path, including the the rightmost directory. Since customed firmware base loading directory for kernel modules is /var/lib/firmware, it means that kernel modules that are expecting their firmware to be located unnder base path (/var/lib/firmware) cannot be run, since current implementation will always contain at leas one additional directory under /var/lib/firmware. This PR changes implementation, so that, if needed, firmware file can be copied directly under the base path (/var/lib/firmware)

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -366,7 +366,7 @@ func MakeLoadCommand(spec kmmv1beta1.ModprobeSpec, modName string) []string {
 	var loadCommand strings.Builder
 
 	if fw := spec.FirmwarePath; fw != "" {
-		fmt.Fprintf(&loadCommand, "cp -r %s %s && ", fw, nodeVarLibFirmwarePath)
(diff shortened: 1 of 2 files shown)


internal/daemonset/daemonset.go
```diff
@@ -415,7 +415,8 @@ func MakeUnloadCommand(spec kmmv1beta1.ModprobeSpec, modName string) []string {
 
 	fwUnloadCommand := ""
 	if fw := spec.FirmwarePath; fw != "" {
-		fwUnloadCommand = fmt.Sprintf(" && cd %s && find |sort -r |xargs -I{} rm -d %s/$(basename %s)/{}", fw, nodeVarLibFirmwarePath, fw)
+		//fwUnloadCommand = fmt.Sprintf(" && cd %s && find |sort -r |xargs -I{} rm -d %s/$(basename %s)/{}", fw, nodeVarLibFirmwarePath, fw)
```
ybettan, line 418: Forgot to remove

PR #306, opened 2022-12-19
Build signer images in CI (#165)

ybettan:
Use UBI base images for a smaller footprint.
Rework the Dockerfile for better caching.

Co-authored-by: Yoni Bettan <yonibettan@gmail.com>

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/257

PR #308, opened 2022-12-19
Move to the kmm.sigs.x-k8s.io API group (#214)

qbarrand:
Move to the x-k8s.io API Group as we do not need a strict API review process.

Fixes #306


ybettan:
I don't think https://github.com/rh-ecosystem-edge/kernel-module-management/pull/306 is related here.
It is supposed to fix https://github.com/openshift/release/pull/34904.

PR #320, opened 2023-01-03
Fixing sign image generation

yevgeny-shnaidman:
This PR does the following:
1) fixing Dockerfile.signimage. moving the final stage image from
   ubi-micro to ubi-minimal, since sign-file executable needs shared
   libraries which are deployed by openssl-libs package. This package
   is not present in the ubi-micro image
2) moving from podman to docker in the Makefile in order
   to be consistent with the rest of the targets
3) updating the skipper.env to allow building sign image with
   a specificied target


chr15p:
I dont understand why we use docker hardcoded rather than setting it to a variable (like we do for other binaries e.g. controller-gen), so I think we should use the $(PODMAN) variable everywhere rather than removing it.

PR #343, opened 2023-01-11
Removing all e2e github workflows but the `in-cluster-build` one. (#224)

ybettan:
This detach the running cluster from its backend, for example running any other cluster than a cluster using the host kernel (such as `minikube` with the `docker` driver or `kind`) and allow us to re-use those tests in other cluster envs.

The following jobs are removed:

* `e2e-two-nodes`
* `e2e-crd-variable`
* `e2e-pre-built-kernel-module`

Those jobs are not testing anything special, one is just testing the selector of the `ds`, the second is just using a regex in the `kernelMapping` flow which IMO should be both tested in UT rather than having their own e2e workflows and the last one is just replicating the second part of the `in-cluster-build`.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/341
/cc @qbarrand @yevgeny-shnaidman @enriquebelarte


ci/module-kmm-ci-build.yaml
```diff
@@ -8,9 +8,13 @@ spec:
     container:
       modprobe:
         moduleName: kmm_ci_a
-      imagePullPolicy: Never
       kernelMappings:
         - regexp: '^.+$'
           containerImage: image-registry.openshift-image-registry.svc:5000/default/kmm-kmod:$KERNEL_FULL_VERSION
+          registryTLS:
+            insecure: true
```
qbarrand, line 15: We do not need this.

ci/prow/e2e-incluster-build
```diff
@@ -12,21 +12,22 @@ oc label node ${NODE} task=kmm-ci
...
-
-sed -e "s/KVER_CHANGEME/${KVER}/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
-oc apply -f module-kmm-ci.yaml
+oc apply -f ci/module-kmm-ci-build.yaml
 
 # Wait for the build pod to be created. `kubectl wait` doesn't support such option,
 # see https://github.com/kubernetes/kubernetes/issues/83242.
 echo "Waiting for the build pod to be created..."
-timeout 5m bash -c 'until oc get pods -o json | jq -er ".items[].metadata.name | select(.? | match(\"build\"))"; do sleep 1; done'
+timeout 10m bash -c 'until oc get pods -o json | jq -er ".items[].metadata.name | select(.? | match(\"build\"))"; do sleep 1; done'
```
qbarrand, line 30: Does it really take 10 minutes to create the build pod? If yes, we have a problem - it should be a matter of seconds.

ci/module-kmm-ci-build.yaml
ybettan: This is just `git` misleading here. Since `module-kmm-ci-build.template.yaml` was renamed to `module-kmm-ci-build.yaml` git shows the diff from the `module-kmm-ci-crd` (probably because it is the closes to this one after the change).

So the `registryTLS` was already present in the module (this PR didn't change it). The only differences in this file (compared to its previous version `module-kmm-ci-build.template.yaml` are:
* removing the build args
* removing the build secret (which was added back later on and will be pull m/s at some point)

ybettan: If you want to validate the change you can check the diff between `module-kmm-ci-build.yaml` in this PR and `module-kmm-ci-build.template.yaml` from the `main` branch.

ybettan: Let me make it simple for you. Here is the real diff:
```
root kernel-module-management-midstream (e2e-cleanup) $ sdiff ci/module-kmm-ci-build.yaml ../module-kmm-ci-build.template.yaml
---                                                             ---
apiVersion: kmm.sigs.x-k8s.io/v1beta1                           apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: Module                                                    kind: Module
metadata:                                                       metadata:
  name: kmm-ci                                                    name: kmm-ci
spec:                                                           spec:
  moduleLoader:                                                   moduleLoader:
    container:                                                      container:
      modprobe:                                                       modprobe:
        moduleName: kmm_ci_a                                            moduleName: kmm_ci_a
      kernelMappings:                                                 kernelMappings:
        - regexp: '^.+$'                                      |         - literal: KVER_CHANGEME
          containerImage: image-registry.openshift-image-regi |           containerImage: image-registry.openshift-image-regi
          registryTLS:                                                    registryTLS:
            insecure: true                                                  insecure: true
          build:                                                          build:
                                                              >             buildArgs:
                                                              >               - name: CI_BUILD_ARG
                                                              >                 value: some-build-arg
                                                              >             secrets:
                                                              >               - name: build-secret
            dockerfileConfigMap:                                            dockerfileConfigMap:
              name: kmm-kmod-dockerfile                                       name: kmm-kmod-dockerfile
  selector:                                                       selector:
    task: kmm-ci                                                    task: kmm-ci

```

ci/prow/e2e-incluster-build
ybettan: No. It should be triggered immediately (it is working locally but not in CI before timeout even after 10m for some reason. Still debugging this one.

ci/module-kmm-ci-build.yaml
qbarrand: Let's anyway remove that line? 🙂

ci/prow/e2e-incluster-build
ybettan: Seems like when I have increased it to 60m for debugging it passed (and failed with 10m) but I don't have timestamps on the logs so I don't know how long it actually took: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/rh-ecosystem-edge_kernel-module-management/343/pull-ci-rh-ecosystem-edge-kernel-module-management-main-e2e-in-cluster-build/1613426413178720256/artifacts/e2e-in-cluster-build/run-test/build-log.txt. I guess that it probably depends on the cluster we get?!

ci/module-kmm-ci-build.yaml
ybettan: It means that we will skip TLS u/s but not m/s. Is this what we want?

qbarrand: Yes. Upstream, we use a plain HTTP registry. Downstream, we have TLS enabled and all the required CAs.

ybettan: Ack. Removed

logs.txt
```diff
@@ -0,0 +1,114 @@
+sh-4.2$ oc logs pod/kmm-operator-controller-manager-5b5dcdf479-7cdp5 -n kmm-operator-system
```
qbarrand, line 1: Please remove this

ybettan: OOps

ybettan: Fixed.

PR #344, opened 2023-01-11
Automatically add the builder service account secrets to the signing

chr15p:
Automatically add the builder service account secrets to the signing process

Both spec.imageRepoSecret and the builder secret are mounted as volumes under /docker_config 
Changes in signimage in another PR means it then walks this tree looking for secrets and the appropriate secret for the repo is used.  This results in both the builder secret and spec.imageRepoSecret being valid places for push and pull credentials to reside.

### DIFF
--- a/internal/sign/job/signer.go
+++ b/internal/sign/job/signer.go
@@ -126,11 +126,24 @@ func (s *signer) MakeJobTemplate(
 		utils.MakeSecretVolumeMount(signConfig.KeySecret, "/signingkey"),
 	}
 
-	args = append(args, "-secretdir", "/docker_config/")
 	imageSecret := mod.Spec.ImageRepoSecret
+	buildImageSecret, err := s.getSAImageRepoSecret(ctx, &mod, constants.OCPBuilderServiceAccountName)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get secrets for service account %s: %v", constants.OCPBuilderServiceAccountName, err)
+	}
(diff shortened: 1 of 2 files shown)


chr15p:
/hold

Waiting on PRs to be merged from upstream

internal/sign/job/signer.go
```diff
@@ -182,6 +190,22 @@ func (s *signer) getHashAnnotationValue(ctx context.Context, privateSecret, publ
...
+
+	err := s.client.Get(ctx, namespacedName, &serviceaccount)
+	if err != nil {
+		return nil
+	}
+
+	if len(serviceaccount.Secrets) == 0 {
+		return nil
+	}
+	return &v1.LocalObjectReference{Name: serviceaccount.Secrets[0].Name}
```
qbarrand, line 206: Don't we want to return the whole slice and mount that?

chr15p: Annoyingly we get ObjectReference's and we need LocalObjectReference's for the secrets so this is a hoop that casts these.  However I was assuming that the serviceaccount only had one secret, so I've changed this so it returns the serviceaccount.Secrets array and we loop through that to create out volumes (casting as we go)

I will add removing the casts to my list of future improvements but it needs changes to functions other used in other places so will need a wider PR.

chr15p:
/unhold

still waiting on #353  as a dependency but that's very much nearly almost there....

PR #350, opened 2023-01-12
Adding back the build-secret to the e2e test. (#232)

ybettan:
The decision to remove it was a mistake. It was confused with
`imageRepoSecret`.

The build arg is supposed to be available to the `build` Dockerfile at
`/run/secrets/build-secret/filename`.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/346

### DIFF
--- a/ci/kmm-kmod-dockerfile.yaml
+++ b/ci/kmm-kmod-dockerfile.yaml
@@ -13,6 +13,8 @@ data:
 
     WORKDIR /usr/src
 
+    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret
+
     RUN git clone https://github.com/rh-ecosystem-edge/kernel-module-management.git
 
     WORKDIR /usr/src/kernel-module-management/ci/kmm-kmod
--- a/ci/prow/e2e-incluster-build
+++ b/ci/prow/e2e-incluster-build
@@ -18,6 +18,9 @@ if oc debug node/${NODE} -- chroot host/ lsmod | grep kmm_ci_a; then
  exit 1
 fi
 
+echo "Create a build secret..."
+oc create secret generic build-secret --from-literal=ci-build-secret=super-secret-value
+
 echo "Add a configmap that contain the kernel module build dockerfile..."
 oc apply -f ci/kmm-kmod-dockerfile.yaml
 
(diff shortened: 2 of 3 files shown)


ybettan:
/hold
Need to rebase it on top of https://github.com/rh-ecosystem-edge/kernel-module-management/pull/343 once merged.

ci/prow/e2e-incluster-build
```diff
@@ -30,10 +33,14 @@ echo "Waiting for the build pod to be created..."
...
 POD_NAME=$(oc get pods -o json | jq -r '.items[].metadata.name | select(.? | match("build"))')
 
-# The build job/pod is deleted once done so we won't be able to get this info in the CI artifacts.
-echo "Print the build logs..."
-# we can't get the logs on a pod that isn't `Running` yet.
+# we can't exec a command nor get the logs on a pod that isn't `Running` yet.
 oc wait pod/${POD_NAME} --for jsonpath='{.status.phase}'=Running --timeout=60s
+
+# Check that the build secret is available to the build pod
+timeout 1m bash -c "until oc exec ${POD_NAME} -- grep super-secret-value /run/secrets/build-secret/ci-build-secret; do sleep 3; done"
```
qbarrand, line 40: I think we should modify the Dockerfile to write the contents of `/run/secrets/build-secret/ci-build-secret` into the built image and grep once the ModuleLoader pod is running.

ybettan:
After talking to @yevgeny-shnaidman the test is failing because there is a real bug in the code.

We are using the wrong object in order to mount the secret to the `build` object.

When we will use the correct object, the secret will always be mounted to the same directory as the Dockerfile (so the build can copy the file into the Dockerfile while building the image).

So the issues are:
* the secret is mounted in a default location aka `/var/run/openshift.io/volumes/build-secret-user-build-volume` instead of what we are specifiying in the code `/run/secrets/secret-name`
* even u/s we have an issues as the secrets are available at /run/secrets but the Dockerfile may be elsewhere and the Dockerfile won't be able to make use of that secret.

FYI @qbarrand

ci/prow/e2e-incluster-build
ybettan: Why not just `RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret && git clone ...`?
All we really want is that the Dockerfile is able to read some secrets before running `git clone` no?!

ci/kmm-kmod-dockerfile.yaml
```diff
@@ -13,7 +13,8 @@ data:
 
     WORKDIR /usr/src
 
-    RUN git clone https://github.com/rh-ecosystem-edge/kernel-module-management.git
+    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret && \
+        git clone https://github.com/rh-ecosystem-edge/kernel-module-management.git
```
qbarrand, line 17:
```suggestion
    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret
    RUN git clone https://github.com/rh-ecosystem-edge/kernel-module-management.git
```

ybettan: I don't mind by why? I would like to think about it as git clone --auth ... so it makes sense to me to put in in the same layer.

qbarrand: It can be used for pretty much anything you want, `git clone` is just an example.
I think having it in the same `RUN` as the clone makes it more confusing.

PR #393, opened 2023-01-24
Bump Kubernetes dependencies (#215)

ybettan:
Move to the new sets.Set type as sets.String is now deprecated. Bump mockgen to a version that supports generics.
Bump golangci-lint.

Co-authored-by: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/389


qbarrand:
/hold
@ybettan OCP 4.12 uses Kubernetes 1.25, we should not upgrade those deps.

ybettan:
@qbarrand Indeed but I wanted to check if it passes CI and it does. Seems like a newer client still work with older clusters.
Do you still think we shouldn't update the API version?

qbarrand:
> Seems like a newer client still work with older clusters.

Some features might break. I do not see the benefit tbh.

PR #404, opened 2023-01-25
Suggest namespace in the bundles

qbarrand:
Install KMM in `openshift-kmm` and KMM-Hub in `openshift-kmm-hub`.


ybettan:
Do we also change the namespace in the non-bundle installation?
```
$ git grep kmm-operator-system
config/default-hub/kustomization.yaml:namespace: kmm-operator-system
config/default/kustomization.yaml:namespace: kmm-operator-system
docs/labs/README.md:Once KMM is installed a new ReplicaSet will be deployed and a new pod running in the `kmm-operator-system` namespace will include the KMM controller.
```

ybettan:
/lgtm
/hold
Unhold if you wish to merge.

PR #418, opened 2023-02-02
Add must-gather as a related image

qbarrand:
This is useful for users to determine which must-gather image must be used to troubleshoot their deployment.

/cc @mresvanis @ybettan

### DIFF
--- a/config/manager-base/manager.yaml
+++ b/config/manager-base/manager.yaml
@@ -46,6 +46,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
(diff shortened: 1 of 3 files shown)


config/manager-base/manager.yaml
```diff
@@ -46,6 +46,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: RELATED_IMAGES_MUST_GATHER
```
ybettan, line 49: Don't we need the same for `manager-hub`?

PR #432, opened 2023-02-12
Refactor mapping and module unification and override (#284)

yevgeny-shnaidman:
Currently we have nultiple definitions (Build, Sign, ContainerImage, TLS options) that can be defined both in a mapping and also in the Module Container (more global). These are checked in mulitple places of the code ( during building, signing, ModuleLoader creation etc'), which make it very confusing to keep notice of,and in addition, makes it hard to add new variable both to KerneMapping and to ModuleLoader container This PR moves all the unifications/overrides into one place, at the beginning of the Reconcile loop, and stores the unified data in the KernelMapping. From that point on the code needs to work only with KernelMapping, without any need to check what is the data in ModuleLoadContainer Flow:
- find kernel mapping relevent for kernel version
- unifify data of the Container and Kernel Mapping into the kernel mapping
- all the sign/build/driver container code is now working only with kernel mapping

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/428


yevgeny-shnaidman:
Upstream cherry-pick: 4bc55e595e2d5c4d46ff3732d272d5d9573c4d27

PR #443, opened 2023-02-14
Simplifying KernelAPI interface (#292)

yevgeny-shnaidman:
Currently KernelAPI consists of 4 functions, which are always used together in a sequence. This redifines the KernelAPI interface to contain only one functions that does everything the previous 4 functions were doing. In addtion a private helper interface is defined, and its function are used by the KernelAPI interface


yevgeny-shnaidman:
Upstream cherry-pick 33708ec025887340d726683f96279c273abce6d6

yevgeny-shnaidman:
Fixes [#443 ](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/441)

PR #444, opened 2023-02-14
ModuleReconciler and ManagedCluster controllers should not requeue on…

yevgeny-shnaidman:
… Build/Sign valid statuses (#296)

Currently both controllers are requeueing the reconciliation event, in case Build or Sign has returned a flag of requeue. This is not necessary, since both controllers are watching the owned Job object. This PR changes this logic so that in case Build or Sign job has returned valid, but not complete status (Created, In Progress, Failed), the reconciliation request is not requeued The changes are:
1) Manager functions of Build, Sign return only the current status of the job and error code.
   No need to return requeueing indication. The statuses returned are defined in the utils packages
2) The boolean return code of HandleBuild and HandleSign now indicates whether the Build or Sign
   jobs habe been successfully completed or not
3) In the reconciliation loop, in case HandleBuild or HandleSign successfull completion indicator
   is false, the flow does not continue to the ModuleLoader creation
4) No matter which if Sign/Build job has completed successfully or not, the reconciliation request
   is not requeued. The request is requeued only in case of an actual error in the flow

### DIFF
--- a/controllers/hub/managedclustermodule_reconciler.go
+++ b/controllers/hub/managedclustermodule_reconciler.go
@@ -110,14 +110,13 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, req ctrl
 		logger := log.FromContext(ctx).WithValues("cluster", cluster.Name)
 		clusterCtx := log.IntoContext(ctx, logger)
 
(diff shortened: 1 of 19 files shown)


yevgeny-shnaidman:
Upstream cherry-pick 65198ea64d69f22fc3e4a65164d907624f1b2fd5

PR #449, opened 2023-02-16
Creating ModuleLoaderData sturcture for Build/Sign/ModuleLoader workf…

yevgeny-shnaidman:
…low (#301)

This PR add the creation of the ModuleLoaderData structure. This struct contains all the data needed for processing the wholw flow of Build/Sign/Module loader per specific Module/KernelMapping.
The flow ass following:
1) when starting processing the Reconcile request( either in Module controller
   or in ManagedCluster controller) all the needed data from Module and KernelMapping
   is merged into a newly created ModuleLoaderData struct
2) all the following layers are working only with this structure, no need to look
   at Module or KernelMapping again.
3) This does not affect the flow of DevicePlugin

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -23,6 +23,7 @@ import (
 
 	buildv1 "github.com/openshift/api/build/v1"
 	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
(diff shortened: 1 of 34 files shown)


yevgeny-shnaidman:
Upstream cherry-pick 244d756a30e633dd6e96164ae388f08828faf1ae

PR #451, opened 2023-02-20
Refactoring ModuleReconciler (#306)

yevgeny-shnaidman:
This PR add an inner helper interface to the ModuleReconciler. Now the reconcile loop almost entirely uses interface functions in its logic, which allows for shorter and saner unit-tests Changes in PR:
1) splitting ModuleReconcile to Reconcile loop and helper interface 2) removing client.Get call from handleDevicePlugin (not needed, will
   be called anyway by controllerutil.CreateOrPatch
3) unit tests rewrite


yevgeny-shnaidman:
Upstream cherry-pick y979333c4125db1767552843b41aad1f9386a0147

ybettan:
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/450
/lgtm

PR #453, opened 2023-02-21
Documentation fix

yevgeny-shnaidman:
Fixing the build Configmap example in the secture_boot.md The key for the Configmap's data should be 'dockerfile' and not 'Dockerfile'


yevgeny-shnaidman:
The fix was already back-ported from the upstream

PR #456, opened 2023-02-21
Changing the metrics flow (#307)

yevgeny-shnaidman:
Up to now, the metric were mainly refelcting the current progress of customer's Module, which was not very useful, since customers do not look at them. This PR changes the metrics so as to supply the info regarding KMMO usage by the customer. The following metrics are gathered: 1) number of Modules deployed on the cluster
2) number of modules with in-cluster build capabilty 3) number of module with in-cluster sign capability 4) number of modules deploying device plugin
5) if the preflight capability is used on the cluster 6) modules that are using Morprobe arg configuration ( raw and non-raw)
   and which configuration they are using

---

Fixes[#452 ]

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -160,7 +160,7 @@ func main() {
 		kernelAPI,
 		metricsAPI,
 		filterAPI,
-		statusupdater.NewModuleStatusUpdater(client, metricsAPI),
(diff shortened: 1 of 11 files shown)


yevgeny-shnaidman:
Upstream cherry-pick ad710cb9ee7f1103263def9e2842be5e6499bd36

PR #459, opened 2023-02-23
Add an enhancement for disconnected support

qbarrand:
Rename the previous enhancement to include an identifier.

/cc @mresvanis @yevgeny-shnaidman @ybettan @chr15p 

Relates to MGMT-13819.


docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
+# Disconnected support for KMM
```
ybettan, line 1: Why do we need identifiers to the file names?

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+The registry is installed and managed by the cluster administrator.
+
+Red Hat offers [documentation](https://docs.openshift.com/container-platform/4.12/installing/disconnected_install/installing-mirroring-installation-images.html#installing-mirroring-installation-images)
+to mirror the OpenShift system iamges on a private registry.
+Under the hood, the tools used:
+- copy the container images from the public registry on the Internet to the private one;
+- create [`ImageContentSourcePolicy`] objects that configure the mirroring on nodes.
+  [`ImageContentSourcePolicy`] objects configure alternate locations for images specified **by digest only**.
+
+KMM itself can be mirrored like any operator with the usual tools.
```
cdvultur, line 19: Wondering if we should not add some note related to DTK , and it's relation to the running RHCOS.

cdvultur: Still, DTK is part of OCP, so it should be there with the release payload..

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Introduction: disconnected in OCP
+
+The term "disconnected clusters" refers to clusters that do not have Internet access and thus cannot download publicly
+available container images.
+To support disconnected clusters, a private registry must be used.
+The registry is installed and managed by the cluster administrator.
+
+Red Hat offers [documentation](https://docs.openshift.com/container-platform/4.12/installing/disconnected_install/installing-mirroring-installation-images.html#installing-mirroring-installation-images)
+to mirror the OpenShift system iamges on a private registry.
```
ybettan, line 13: sed/iamges/images

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+In those cases, support for mirrored registries as configured in [`ImageContentSourcePolicy`] objects must be
+implemented in the operator code.
+
+## In KMM
+
+The scenarios where the operator or one of its components might need to build an image are listed below:
+
+| Scenario                                                                            | Handled by OCP |  Work needed  |
+|-------------------------------------------------------------------------------------|:--------------:|:-------------:|
+| When the `Module` reconciler checks if the output image already exists              |       ❌        | ❌<sup>1</sup> |
```
ybettan, line 32: This part is not clear to me, don't we have needed work here? We need extra code to check the mirrored image instead of the public one.

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+| When the `PreflightValidation` reconciler checks if the output image already exists |       ❌        | ❌<sup>1</sup> |
```
ybettan, line 33: Same here.

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+| When pulling the base image (`FROM` in the `Dockerfile`)                            | ✅<sup>2</sup>  |       ❌       |
```
ybettan, line 34: Maybe worth mentioning that the registry auth secret will need to fit the mirrored registry and not the public one.

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+| When pulling the unsigned image                                                     |       ❌        | ✅<sup>3</sup> |
```
ybettan, line 35: Isn't pulling the unsigned image done via a job in OCP? If this is the case then OCP should handle the pull for us isn't it?

docs/enhancements/0001_disconnected.md
qbarrand: If build and / or sign is specified, the image cannot be specified by digest, so no mirroring is possible.

docs/enhancements/0001_disconnected.md
qbarrand: It does not - we pull from **inside** the job and don't use any mechanism that would have that automated.

docs/enhancements/0001_disconnected.md
```diff
@@ -0,0 +1,93 @@
...
+available container images.
+To support disconnected clusters, a private registry must be used.
+The registry is installed and managed by the cluster administrator.
+
+Red Hat offers [documentation](https://docs.openshift.com/container-platform/4.12/installing/disconnected_install/installing-mirroring-installation-images.html#installing-mirroring-installation-images)
+to mirror the OpenShift system iamges on a private registry.
+Under the hood, the tools used:
+- copy the container images from the public registry on the Internet to the private one;
+- create [`ImageContentSourcePolicy`] objects that configure the mirroring on nodes.
+  [`ImageContentSourcePolicy`] objects configure alternate locations for images specified **by digest only**.
```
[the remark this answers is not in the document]
qbarrand: It's also capable of replacing names with a path, e.g. `quay.io/some-user` --> `quay.io/some-other-user`.

PR #471, opened 2023-03-07
Make the deployment in CI optional

qbarrand:
To allow the installation in CI via the bundle, introduce `SKIP_MAKE_DEPLOY` which skips `make deploy` when `true`.

/cc @enriquebelarte

### DIFF
--- a/ci/prow/e2e-incluster-build
+++ b/ci/prow/e2e-incluster-build
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+readonly SKIP_MAKE_DEPLOY="${SKIP_MAKE_DEPLOY:-false}"
 POD_NAME=''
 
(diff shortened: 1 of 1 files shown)


ci/prow/e2e-incluster-build
```diff
@@ -19,8 +20,12 @@ export KVER=$(oc debug node/${NODE} -- uname -r)
 echo "Label a node to match selector in Module afterwards..."
 oc label node ${NODE} task=kmm-ci
 
-echo "Deploy KMMO..."
-make deploy
+if [ "$SKIP_MAKE_DEPLOY" == true ]; then
+  echo Skipping deployment
```
enriquebelarte, line 24: Maybe double quotes missing in `echo Skipping deployment` ?

Issue #410, opened 2023-01-29
Day1 kernel module support via KMM in OCP


yevgeny-shnaidman:
# Issue Summary
Currently KMM supports only Day2 operations: loading/replacing/upgrading kernel modules only after full installation of the OCP cluster. Allowing some kind of Day1 support ( loading kernel module prior to full cluster installation) will increase the usability of KMM

# Proposed solution
The solution is rendered on the level of root FS, and not on the level of initrmfs, using the MCO, ignition configuration and the same driver containers as used by KMM
- 2 systemd services and 2 executable will be deployed via MCO
- both services will be one-shot services
- one service (Install Service) will be run before Network configuration and the second one (Pull Image Service) after network configuration
- the first service will check the presence of the configured Driver Container Image on the node, and in case it is present, will install the the kernel module defined in the image
- the second service will check the presence of the DriverContainer image in the node, and in case it is not present, will pull it onto the node and reboot the node immediately

<details>
<summary>Example MachineConfig</summary>

```yaml
apiVersion: machineconfiguration.openshift.io/v
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: sfc
  name: replace-sfc
spec:
  config:
    ignition:
      version: 3.2.0
    systemd:
      units:
        - contents: |
            [Unit]
            Description=Replace in-tree sfc driver with oot sfc driver
            Before=network-pre.target
            Wants=network-pre.target
            DefaultDependencies=no
            [Service]
            User=root
            Type=oneshot
            TimeoutSec=10
            ExecStartPre=ls /usr/local/bin
            ExecStart=/usr/local/bin/replace-sfc-driver.sh
            PrivateTmp=yes
            RemainAfterExit=no
            TimeoutSec=60
            [Install]
            WantedBy=multi-user.target
          enabled: true
          name: "replace-sfc.service"
        - contents: |
            [Unit]
            Description=Pull oot sfc driver container
            After=network-online.target
            Wants=network-online.target
            DefaultDependencies=no
            [Service]
            User=root
            Type=oneshot
            ExecStart=/usr/local/bin/pull-sfc-driver.sh
            PrivateTmp=yes
            RemainAfterExit=no
            TimeoutSec=900
            [Install]
            WantedBy=multi-user.target
          enabled: true
          name: "pull-sfc-image.service"
        - enabled: false
          mask: true
          name: crio-wipe.service
    storage:
      files:
[...]

PR #474, opened 2023-03-15
Adding MC CR generation external package for KMM Day1 support

yevgeny-shnaidman:
This PR add am external package that will allow user to generate an MC CR yaml file. When applied , the MCO will cause a kernel module to be replaced with a customer's image before NetworkManager service starts running ( basically Day1). Pre-requisite: at least one valid network interface should be running on the node without MCO interference
The produced MC CR defines 2 systemd services:
1) pull service: executes after network interface (NM) is configured, and is responsible for pull
    kernel module image, based on DTK, into the node. Once successfull, it reboots the node
2) replace kernel module service: executes before NM is configured ( no network). In case the image is already
   present on the node, it will remove the in-tree kernel module, and will replce it with the kernel module
   stored in the DTK-based image

The input to function that produce MC CR yaml is:
1) MC CR name - the name that will be set into the MC.name
2) MCP reference - reference to the MachineConfigPool, which will define which nodes this MC CR will affect.
3) image name - DriverContainer image, that will contain the needed kerne module driver 4) the name of the kernel driver - which kernel driver to remove/insert

### DIFF
--- a/pkg/mcproducer/mcproducer.go
+++ b/pkg/mcproducer/mcproducer.go
@@ -1,0 +1,62 @@
+package mcproducer
+
+import (
+	"bytes"
+	"embed"
+	"encoding/base64"
+	"fmt"
+	"text/template"
+
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+//go:embed templates
+var templateFS embed.FS
+
+func ProduceMachineConfig(machineConfigName, machineConfigPoolRef, kernelModuleImage, kernelModuleName string) (string, error) {
+	tag, err := name.NewTag(kernelModuleImage)
+	if err != nil {
+		return "", fmt.Errorf("invalid kernelModuleImage %s, input should be repo:tag format: %v", kernelModuleImage, err)
+	}
+
+	templateParams := map[string]interface{}{
+		"Image":                tag.Repository.Name(),
+		"Tag":                  tag.Identifier(),
+		"KernelModule":         kernelModuleName,
+		"MachineConfigPoolRef": machineConfigPoolRef,
+		"MachineConfigName":    machineConfigName,
+	}
+
+	replaceKernelModuleScript, err := generateTemplate(templateFS, "templates/replace-kernel-module.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	pullKernelModuleScript, err := generateTemplate(templateFS, "templates/pull-image.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	replaceScriptBase64 := base64.StdEncoding.EncodeToString([]byte(replaceKernelModuleScript))
+	pullScriptBase64 := base64.StdEncoding.EncodeToString([]byte(pullKernelModuleScript))
+
+	templateParams["ReplaceInTreeDriverContents"] = replaceScriptBase64
+	templateParams["PullKernelModuleContents"] = pullScriptBase64
+
+	machineConfigYAMLString, err := generateTemplate(templateFS, "templates/machine-config.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	return machineConfigYAMLString, nil
+}
+
+func generateTemplate(fsys embed.FS, fileName string, params map[string]interface{}) (string, error) {
+	tmpl, err := template.ParseFS(fsys, fileName)
+	if err != nil {
+		return "", fmt.Errorf("failed to prepare template from file %s: %v", fileName, err)
+	}
+	buf := &bytes.Buffer{}
+	err = tmpl.Execute(buf, params)
+	if err != nil {
+		return "", fmt.Errorf("failed to execute template parsing for file %s: %v", fileName, err)
+	}
+	return buf.String(), nil
+}
--- a/pkg/mcproducer/mcproducer_test.go
+++ b/pkg/mcproducer/mcproducer_test.go
@@ -1,0 +1,40 @@
+package mcproducer
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"os"
+)
+
+var _ = Describe("ProduceMachineConfig", func() {
+	const (
+		name             = "name"
+		mcpRef           = "mcpRef"
+		kernelModuleName = "testKernelModuleName"
+	)
+
+	It("image name format is invalid", func() {
+		imageName := "quay.io/project/repo@sha256:1f5f1ae25db67aa82707e1b1dc96c8a53ef7094f320b7eeaef12be9a13fa251d"
+
+		res, err := ProduceMachineConfig(name, mcpRef, imageName, kernelModuleName)
+
+		Expect(err).To(HaveOccurred())
+		Expect(res).To(Equal(""))
+
+	})
+
+	// unit performs verification of the MC output vs the manually created
+	// machineconfig-test.yaml. This yaml is created by manually running bas64
+	// encoding on testdata/pull-image-test.sh and testdata/replace-kernel-module-test.sh
+	// and setting the values into the testdata/machineconfig-test.yaml
+	It("verify correct mco output", func() {
+		imageName := "quay.io/project/repo:some-tag12"
+
+		res, err := ProduceMachineConfig(name, mcpRef, imageName, kernelModuleName)
+
+		Expect(err).ToNot(HaveOccurred())
+		expectedRes, err := os.ReadFile("testdata/machineconfig-test.yaml")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(res).To(Equal(string(expectedRes)))
+	})
+})
--- a/pkg/mcproducer/suite_test.go
+++ b/pkg/mcproducer/suite_test.go
@@ -1,0 +1,23 @@
+package mcproducer
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/test"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var scheme *runtime.Scheme
+
+func TestSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	var err error
+
+	scheme, err = test.TestScheme()
+	Expect(err).NotTo(HaveOccurred())
+
+	RunSpecs(t, "Job Suite")
+}
--- a/pkg/mcproducer/templates/machine-config.gotmpl
+++ b/pkg/mcproducer/templates/machine-config.gotmpl
@@ -1,0 +1,67 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: {{.MachineConfigPoolRef}}
+  name: {{.MachineConfigName}}
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=Replace in-tree kernel module with oot kernel module
+            Before=network-pre.target
+            Wants=network-pre.target
+            DefaultDependencies=no
+            [Service]
+            User=root
+            Type=oneshot
+            TimeoutSec=10
+            ExecStartPre=ls /usr/local/bin
+            ExecStart=/usr/local/bin/replace-kernel-module.sh
+            PrivateTmp=yes
+            RemainAfterExit=no
+            TimeoutSec=60
+            [Install]
+            WantedBy=multi-user.target
+          enabled: true
+          name: "replace-kernel-module.service"
+        - contents: |
+            [Unit]
+            Description=Pull oot kernel module image
+            After=network-online.target
+            Wants=network-online.target
+            DefaultDependencies=no
+            [Service]
+            User=root
+            Type=oneshot
+            ExecStart=/usr/local/bin/pull-kernel-module-image.sh
+            PrivateTmp=yes
+            RemainAfterExit=no
+            TimeoutSec=900
+            [Install]
+            WantedBy=multi-user.target
+          enabled: true
+          name: "pull-kernel-module-image.service"
+        - enabled: false
+          mask: true
+          name: crio-wipe.service
+    storage:
+      files:
+        - path: "/usr/local/bin/replace-kernel-module.sh"
+          mode: 511
+          overwrite: true
+          user:
+            name: "root"
+          contents:
+            source: "data:text/plain;base64,{{.ReplaceInTreeDriverContents}}"
+        - path: "/usr/local/bin/pull-kernel-module-image.sh"
+          mode: 493
+          overwrite: true
+          user:
+            name: "root"
+          contents:
+            source: "data:text/plain;base64,{{.PullKernelModuleContents}}"
--- a/pkg/mcproducer/templates/pull-image.gotmpl
+++ b/pkg/mcproducer/templates/pull-image.gotmpl
@@ -1,0 +1,13 @@
+#!/bin/bash
+if podman image list | grep {{.Image}} | grep -q {{.Tag}}; then
+    echo "Image {{.Image}}:{{.Tag}} found in the local registry.Nothing to do"
+else
+    echo "Image {{.Image}}:{{.Tag}} not found in the local registry, pulling"
+    podman pull --authfile /var/lib/kubelet/config.json {{.Image}}:{{.Tag}}
+    if [ $? -eq 0 ]; then
+        echo "Image {{.Image}}:{{.Tag}} has been successfully pulled, rebooting.."
+        reboot
+    else
+        echo "Failed to pull image {{.Image}}:{{.Tag}}"
+    fi
+fi
--- a/pkg/mcproducer/templates/replace-kernel-module.gotmpl
+++ b/pkg/mcproducer/templates/replace-kernel-module.gotmpl
@@ -1,0 +1,20 @@
+#!/bin/bash
+echo "before checking podman images"
+if podman images list | grep {{.Image}} | grep -q {{.Tag}}; then
+    echo "Image {{.Image}}:{{.Tag}} found in the local registry, removing in-tree kernel module"
+    podman run --privileged --entrypoint modprobe {{.Image}}:{{.Tag}} -rd /opt {{.KernelModule}}
+    if [ $? -eq 0 ]; then
+            echo "Succesffully removed the in-tree kernel module {{.KernelModule}}"
+    else
+            echo "failed to remove in-tree kernel module {{.KernelModule}}"
+    fi
+    echo "Running container image to insert the oot kernel module {{.KernelModule}}"
+    podman run --privileged --entrypoint modprobe {{.Image}}:{{.Tag}} -d /opt {{.KernelModule}}
+    if [ $? -eq 0 ]; then
+            echo "OOT kernel module {{.KernelModule}} is inserted"
+    else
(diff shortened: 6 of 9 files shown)


ybettan:
Part of https://github.com/rh-ecosystem-edge/kernel-module-management/issues/410

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+          contents:
+            source: "data:text/plain;base64,$REPLACE_IN_TREE_DRIVER_CONTENTS"
+        - path: "/usr/local/bin/pull-kernel-module-image.sh"
+          mode: 493
+          overwrite: true
+          user:
+            name: "root"
+          contents:
+            source: "data:text/plain;base64,$PULL_OOT_KERNEL_MODULE_CONTENTS"
+`
```
ybettan, line 110: How about we create some data files instead of hardcoding them?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/base64"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/utils"
+)
+
+const replaceKernelModuleScriptTemplate = `#!/bin/bash
+echo "before checking podman images"
+if podman images | grep $IMAGE | grep -q $TAG; then
```
ybettan, line 13:
```suggestion
if podman images | grep -q ${IMAGE}:${TAG}; then
```

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+    echo "Image $IMAGE:$TAG found in the local registry, removing in-tree kernel module"
+    podman run --privileged --entrypoint modprobe $IMAGE:$TAG -rd /opt $KERNEL_MODULE
```
ybettan, line 15:
```suggestion
    echo "Image ${IMAGE}:${TAG} found in the local registry, removing in-tree kernel module"
    podman run --privileged --entrypoint modprobe ${IMAGE}:${TAG} -rd /opt $KERNEL_MODULE
```

Just to be on the safe side.

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+    if [ $? -eq 0 ]; then
+            echo "Succesffully removed the in-tree kernel module $KERNEL_MODULE"
+    else
+            echo "failed to remove in-tree kernel module $KERNEL_MODULE"
+    fi
+    echo "Running container image to insert the oot kernel module $KERNEL_MODULE"
+    podman run --privileged --entrypoint modprobe $IMAGE:$TAG -d /opt $KERNEL_MODULE
```
ybettan, line 22: Same comment as before.

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+    echo "OOT kernel module $KERNEL_MODULE is inserted"
```
ybettan, line 23: Can't this command fail?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+else
+   echo "Image $IMAGE:$TAG is not present in local registry, will try after reboot"
+fi
+`
+const pullKernelModuleScriptTemplate = `#!/bin/bash
+if podman image list | grep $IMAGE | grep -q $TAG; then
```
ybettan, line 29: Can we use `podman images` or use `podman image list` in the previous script as well for consistency?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+else
+    echo "Image $IMAGE:$TAG not found in the local registry, pulling"
+    podman pull --authfile /var/lib/kubelet/config.json $IMAGE:$TAG
+    if [ $? -eq 0 ]; then
+        echo "Image $IMAGE:$TAG has been successfully pulled, rebooting.."
+        reboot
+    else
+        echo "Failed to pull image $IMAGE:$TAG"
+    fi
+fi
```
ybettan, line 40: Can we use ${IMAGE}:${TAG}?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+            [Unit]
+            Description=Replace in-tree kernel module with oot kernel module
+            Before=network-pre.target
+            Wants=network-pre.target
+            DefaultDependencies=no
+            [Service]
+            User=root
+            Type=oneshot
+            TimeoutSec=10
+            ExecStartPre=ls /usr/local/bin
```
ybettan, line 65: Why do we need it?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+            PrivateTmp=yes
+            RemainAfterExit=no
+            TimeoutSec=900
+            [Install]
+            WantedBy=multi-user.target
+          enabled: true
+          name: "pull-kernel-module-image.service"
+        - enabled: false
+          mask: true
+          name: crio-wipe.service
```
ybettan, line 93: What is this service and why it is needed?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+        - path: "/usr/local/bin/pull-kernel-module-image.sh"
+          mode: 493
+          overwrite: true
+          user:
+            name: "root"
+          contents:
+            source: "data:text/plain;base64,$PULL_OOT_KERNEL_MODULE_CONTENTS"
+`
+
+func ProduceMCO(machineConfigName, machineConfigPoolRef, kernelModuleImage, kernelModuleName string) (string, error) {
```
ybettan, line 112:
```suggestion
func ProduceMachineConfig(machineConfigName, machineConfigPoolRef, kernelModuleImage, kernelModuleName string) (string, error) {
```

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", fmt.Errorf("invalid kernelModuleImage %s, input should be repo:tag format: %v", kernelModuleImage, err)
+	}
+
+	scriptsVars := []string{
+		"IMAGE=" + tag.Repository.Name(),
+		"TAG=" + tag.Identifier(),
+		"KERNEL_MODULE=" + kernelModuleName,
+	}
+
+	generatedScripts, err := utils.ReplaceInTemplates(scriptsVars, replaceKernelModuleScriptTemplate, pullKernelModuleScriptTemplate)
```
ybettan, line 124: What do we have such function in our code? `ReplaceInTemplates`.
Why don't we use [golang tempaltes](https://pkg.go.dev/text/template) for that?

pkg/mcoproducer/testdata/mco-test.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: machineconfiguration.openshift.io/v1
```
ybettan, line 1: Can we rename this file to `machineconfig.yaml`?

pkg/mcoproducer/testdata/pull-driver-test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+if podman image list | grep quay.io/project/repo | grep -q some-tag12; then
```
ybettan, line 2: Let's use `podman image list` or `podman images` everywhere.

pkg/mcoproducer/testdata/pull-driver-test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
ybettan, line 1: I think we should delete those files (this and the next one) and instead use the templates script written.

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: and then reading them? I prefer this way, where you can see all the input in one file ( this is the same way we did in AI with all ignition templates.

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: Agree regarding parenthesis, but i don't agree regarding entrypoint: we don't know how that container was defined, and in case it does contain COMMAND or ENTRYPOINT directive we need to explicitly override it

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: same comment as above

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: mainly for debugging: it does not cost much, but it will allow us to see the contents of the /usr/local/bin directory in case something is not working

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: this service is removing unused images from the node, that was not pulled by crio/kubelet

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: we already have utils package that deals with replacing the env variable in KernelMapping and Dockerfile. So instead of using golang templates, i am using already existing package

pkg/mcoproducer/testdata/mco-test.yaml
yevgeny-shnaidman: renaming it to machibeconfig-test.yaml

pkg/mcoproducer/mcoproducer.go
qbarrand: I think we should go with Go templates as they are part of the standard library and can be parsed only once.

romfreiman:
It generates MC (machine config), not MCO (machine config operator :))

pkg/mcoproducer/templates/pull-image.gotmpl
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+if podman image list | grep {{.Image}} | grep -q {{.Tag}}; then
+    echo "Image {{.Image}}:{{.Tag}} found in the local registry.Nothing to do"
+else
+    echo "Image {{.Image}}:{{.Tag}} not found in the local registry, pulling"
+    podman pull --authfile /var/lib/kubelet/config.json {{.Image}}:{{.Tag}}
```
romfreiman, line 6: what will happen in disconnected? icsp?

yevgeny-shnaidman:
> It generates MC (machine config), not MCO (machine config operator :))

fixed

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	machineConfigYAMLString, err := generateTemplates(templateFS, "templates/machine-config.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	return machineConfigYAMLString, nil
+}
+
+func generateTemplates(fsys embed.FS, fileName string, params map[string]interface{}) (string, error) {
+	tmpl, err := template.ParseFS(fsys, fileName)
```
qbarrand, line 52: Could we parse templates only once?

yevgeny-shnaidman: No, because we need to parse the scripts parts first, encode them 64, and that will be an input for MC CR templating

pkg/mcoproducer/mcoproducer.go
ybettan: So this is just a GC for the images that the new service manually pull?

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
+package mcoproducer
```
ybettan, line 1:
```suggestion
package mcproducer
```

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
...
+		"Tag":                  tag.Identifier(),
+		"KernelModule":         kernelModuleName,
+		"MachineConfigPoolRef": machineConfigPoolRef,
+		"MachineConfigName":    machineConfigName,
+	}
+
+	replaceKernelModuleScript, err := generateTemplates(templateFS, "templates/replace-kernel-module.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
```
ybettan, line 33:
```suggestion
        const replaceKernelModuleTmpl = "templates/replace-kernel-module.gotmpl"
	replaceKernelModuleScript, err := generateTemplates(templateFS, replaceKernelModuleTmpl, templateParams)
	if err != nil {
		return "", fmt.Errorf("couldn't generate %s template: %v", replaceKernelModuleTmpl, err)
	}
```

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
...
+	pullKernelModuleScript, err := generateTemplates(templateFS, "templates/pull-image.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
```
ybettan, line 37: Same comment as before

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
...
+	templateParams["PullKernelModuleContents"] = pullScriptBase64
+
+	machineConfigYAMLString, err := generateTemplates(templateFS, "templates/machine-config.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	return machineConfigYAMLString, nil
+}
+
+func generateTemplates(fsys embed.FS, fileName string, params map[string]interface{}) (string, error) {
```
ybettan, line 51:
```suggestion
func generateTemplate(fsys embed.FS, fileName string, params map[string]interface{}) (string, error) {
```

pkg/mcoproducer/testdata/pull-image-test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
ybettan, line 1: We are following the no-extension convention for all our scripts. Can we stay consistent with it?
Example: https://github.com/kubernetes-sigs/kernel-module-management/tree/main/ci/prow.

pkg/mcoproducer/testdata/replace-kernel-module-test.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
ybettan, line 1: We are following the no-extension convention for all our scripts. Can we stay consistent with it?
Example: https://github.com/kubernetes-sigs/kernel-module-management/tree/main/ci/prow.

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: i don't see the need to return a formatted error here, the returned error is pretty much explanatory

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: same comment as above

pkg/mcoproducer/mcoproducer.go
```diff
@@ -0,0 +1,62 @@
...
+	replaceScriptBase64 := base64.StdEncoding.EncodeToString([]byte(replaceKernelModuleScript))
+	pullScriptBase64 := base64.StdEncoding.EncodeToString([]byte(pullKernelModuleScript))
+
+	templateParams["ReplaceInTreeDriverContents"] = replaceScriptBase64
+	templateParams["PullKernelModuleContents"] = pullScriptBase64
+
+	machineConfigYAMLString, err := generateTemplates(templateFS, "templates/machine-config.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: same comment as above

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: fixed

pkg/mcoproducer/testdata/pull-image-test.sh
yevgeny-shnaidman: It is not a script in the sense that it is being executed, it is an input to the unit test. it does not have executable permissions. I gave it an .sh suffix, so that it would be understood from the name that this is a test file for the script being used in the MC CR

pkg/mcoproducer/testdata/replace-kernel-module-test.sh
yevgeny-shnaidman: same comment as above

pkg/mcoproducer/mcoproducer.go
ybettan: It might be. The main benefit IMO with extending the error is that it gives you in the logs the exact last line of *our* code that fails. I find it super handy when debugging but if you insist on not having it then we can leave it as is.

pkg/mcoproducer/mcoproducer.go
ybettan: Answered above.

pkg/mcoproducer/mcoproducer.go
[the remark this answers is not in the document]
ybettan: Same as above as well.

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: but current code will give you the exact line of code that fails, since the error in the generateTemplate function will return the precise location, explanation and error

pkg/mcoproducer/mcoproducer.go
yevgeny-shnaidman: also answered above

pkg/mcproducer/mcproducer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	replaceKernelModuleScript, err := generateTemplate(templateFS, "templates/replace-kernel-module.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	pullKernelModuleScript, err := generateTemplate(templateFS, "templates/pull-image.gotmpl", templateParams)
+	if err != nil {
+		return "", err
+	}
+	replaceScriptBase64 := base64.StdEncoding.EncodeToString([]byte(replaceKernelModuleScript))
```
qbarrand, line 38: I think we should instead use [`base64.NewEncoder`](https://pkg.go.dev/encoding/base64#NewEncoder) and use it as the template output - less copies.

qbarrand:
Follow-up in https://github.com/rh-ecosystem-edge/kernel-module-management/pull/477

PR #483, opened 2023-03-21
Adding Version field to Module CRD (#327)

yevgeny-shnaidman:
Version field will be used in the ordered upgrade
scenario. It signifies the current version of the kenrel module. It should be updated together with the new image in the appropriate kernel mapping. Its update will start the process of the ordered upgrade


yevgeny-shnaidman:
Upstream cherry-pick ba37cffd59dcbc5d311a8a864e17fb84af141809

ybettan:
This PR is a duplication of https://github.com/rh-ecosystem-edge/kernel-module-management/pull/482

PR #486, opened 2023-03-24
Use publicly available images

qbarrand:
Move from `registry.redhat.io` to `registry.access.redhat.com`.


qbarrand:
Requires https://github.com/openshift/release/pull/37652

PR #494, opened 2023-04-16
Add support for digest format images for day1

yevgeny-shnaidman:
This PR does the following:
1) add support for input image to be in digest format and
   not only in tag format
2) updating unit-test
3) adding validation for input image format


pkg/mcproducer/mcproducer.go
```diff
@@ -75,3 +74,15 @@ func executeIntoBase64(tmpl *template.Template, params map[string]any) (string,
...
+func verifyImageFormat(image string) error {
+	_, err := name.NewDigest(image)
+	if err == nil {
+		return nil
+	}
+	_, err = name.NewTag(image)
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("invalid image %s, input should be either in tag or digest format: %v", image, err)
```
ybettan, line 87: If the image is using a bad digest format then the first `if` will contain the error message that will then be overwritten by the second if (cause a digest is obviously not a tag so we will get another error).

yevgeny-shnaidman: fixed

PR #508, opened 2023-04-26
New feature: enable validation webhooks

qbarrand:
- Scaffolding using the operator-sdk
- Adapted the code just enabling Validation webhooks for Create and Update verbs.
- Implement custom logic that validates the module's specs.

/cc @ybettan @yevgeny-shnaidman @mresvanis


config/default/kustomization.yaml
```diff
@@ -17,49 +17,13 @@ bases:
...
+# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
+- ../webhook
 
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
-# crd/kustomization.yaml
-#- manager_webhook_patch.yaml
+# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
+patchesStrategicMerge:
+  - manager_webhook_patch.yaml
+  - webhookcainjection_patch.yaml
```
ybettan, line 26: Do we need those? Seems like they are related to the certificate.

config/default/kustomization.yaml
```diff
@@ -17,49 +17,13 @@ bases:
...
-#    name: webhook-service
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: SERVICE_NAME
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
+configurations:
+  - kustomizeconfig.yaml
```
ybettan, line 29: This is different than what we have u/s. What is the difference here? Where is the `SERVICE_NAME` define for example?

config/default/kustomizeconfig.yaml
```diff
@@ -0,0 +1,4 @@
+namePrefix:
+- kind: Deployment
+  name: controller-manager
+  path: spec/template/spec/volumes/secret/secretName
```
ybettan, line 4: What is this part doing?

config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
```
ybettan, line 1: Should we remove this m/s?

config/default/webhookcainjection_patch.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
ybettan, line 1: Should we remove this m/s?

config/default/kustomization.yaml
ybettan: So, after discussing it, OCP only contains an equivalent to cert-manager and instead of applying `Certificate` objects we just add the `service.beta.openshift.io/serving-cert-secret-name: <secret name>` annotation to the webhook service.

We still need to mount the secret to the webhook pod (KMM pod) so those are still needed here as well.

config/default/kustomization.yaml
ybettan: This is needed in order to prefix the secret name which is set by an annotation and not an object so kustomize won't prefix it.

We have added an [issue](https://github.com/kubernetes-sigs/kernel-module-management/issues/382) u/s to make sure we are prefixing it u/s as well using the correct tools.

config/default/kustomizeconfig.yaml
ybettan: See https://github.com/rh-ecosystem-edge/kernel-module-management/pull/508/files#r1178889583

config/default/manager_webhook_patch.yaml
ybettan: See https://github.com/rh-ecosystem-edge/kernel-module-management/pull/508/files#r1178887591

config/default/webhookcainjection_patch.yaml
ybettan: See https://github.com/rh-ecosystem-edge/kernel-module-management/pull/508/files#r1178887591

PR #512, opened 2023-04-27
Fix ansible dependency on README.md

enriquebelarte:
Package ansible does not include ansible-playbook. It was deprecated in favour of package ansible-core which does include the necessary binary.


enriquebelarte:
Fixes https://issues.redhat.com/browse/MGMT-14439

PR #513, opened 2023-04-27
Remove bash verbose traces

enriquebelarte:


enriquebelarte:
Fixes https://issues.redhat.com/browse/MGMT-14438

tools/sro2kmm/sro2kmm
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 #
-set -ex
```
qbarrand, line 3: Don't we want to keep `-e`? How about `-u`?

enriquebelarte: Done.

PR #540, opened 2023-05-02
In-tree kernel module replacement (#354)

yevgeny-shnaidman:
This PR implements in-tree kernel module replacement per user request:
1) Add a boolean flag to KernelMapping and to Container definition
   In case KernelMapping's field is defined, it takes precedence over
   Container field
2) In case the boolean flag is defined, the PostStart command will now
   first try to remove in-tree kernel module, and only after that will try
   to insert the kernel module stored in the container image
3) unit-test


qbarrand:
/lgtm

Closes #535

PR #573, opened 2023-05-11
Expand and replace spoke kernel mappings before applying ManifestWorks

mresvanis:
This change adds a ModuleSpec processing step before the Hub reconciler applies the respective ManifestWork CRs, which in turn apply the respective Module CRs on the Spokes.

This step includes parsing the
ManagedClusterModule.Spec.ModuleSpec.ModuleLoader.Container.KernelMappings, in order to replace them with literal ones and define the respective container images with their digests (instead of their tags). The latter is necessary for the Spokes to be able to pull the ModuleLoader container images also in a disconnected environment (where registry images are mirrored and digests are necessary to fetch the mirrored images successfully).

This change also ensures that kernel mappings are processed and defined in sorted order every time (by sorting the kernel versions found in each Spoke cluster's ClusterClaims), in order to not trigger additional reconciliation loops on the Spokes just because the kernel mappings changed ordering and not in any other way.

In order to ensure that we don't overwhelm the registry when fetching the image digests, a generic cache was added. For example, if we have 100 targeted Spokes, all with the same kernel version, KMM Hub will fetch the image digest once and cache it, so that the rest 99 ManifestWorks will leverage the latter. This cache items expire after 10 minutes and the expired items are automatically cleaned up every 10 minutes.

Upstream-Commit: 503fe21945fba96183ad24faf5c89dd34d9f6bf9
Upstream-Commit: d44e89d7a791aa3d152a7f0b7ebdcf065d4b3a21

Fixes: #571

[MGMT-14451](https://issues.redhat.com/browse/MGMT-14451)


qbarrand:
/lgtm
/hold
@mresvanis is testing the feature on OpenShift.

PR #579, opened 2023-05-17
Upgrade to Go 1.19 and ubi9

qbarrand:
Make Skipper install a more recent kustomize.

Requires https://github.com/openshift/release/pull/39411

/cc @mresvanis @yevgeny-shnaidman


Dockerfile.signimage
```diff
@@ -1,9 +1,9 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as ksource
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2 as ksource
 
 # install the package that will contain the sign utilities
 RUN ["microdnf", "install", "kernel-devel"]
```
ybettan, line 4: Seems like this package doesn't exist in 9.2 anymore.

PR #587, opened 2023-05-23
Implementing build garbage collector

yevgeny-shnaidman:
This PR implements garbage collection for all OCP build created per specific Module, and whose PhaseStatus is Completed

### DIFF
--- a/internal/build/ocpbuild/manager.go
+++ b/internal/build/ocpbuild/manager.go
@@ -43,9 +43,22 @@ func NewManager(
 }
 
 func (m *manager) GarbageCollect(ctx context.Context, modName, namespace string, owner metav1.Object) ([]string, error) {
+	moduleBuilds, err := m.ocpBuildsHelper.GetModuleBuilds(ctx, modName, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get OCP buils for module %s: %v", modName, err)
+	}
 
-	//Garbage Collection noti (yet) implemented for Build
-	return nil, nil
+	deleteNames := make([]string, 0, len(moduleBuilds))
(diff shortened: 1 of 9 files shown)


ybettan:
/lgtm
/hold
@qbarrand unhold if you are happy to.

internal/build/ocpbuild/manager.go
```diff
@@ -43,9 +43,22 @@ func NewManager(
 }
 
 func (m *manager) GarbageCollect(ctx context.Context, modName, namespace string, owner metav1.Object) ([]string, error) {
```
qbarrand, line 45: WDYT about using something like
```go
	fieldOpt := client.MatchingFields(
		map[string]string{".status.phase": string(buildv1.BuildPhaseComplete)},
	)

	labelOpt := client.MatchingLabels(map[string]string{"key": "value"})

	if err := m.client.DeleteAllOf(ctx, &buildv1.Build{}, fieldOpt, labelOpt); err != nil {
		return err
	}
```

qbarrand:
@yevgeny-shnaidman do you intend to implement it for signing separately?

internal/build/ocpbuild/manager.go
yevgeny-shnaidman: actually, it is good, but then it means that we cannot know what we have deleted and we won't be able to log it (which is think it pretty beneficial for debugging). I suggest that we postpone your suggestion for a later commit, which will start from the upstream: we are going to change the GarbageCollect interface, since we don't really need owner parameter. We can discuss your suggestion in that PR. WDYT?

ybettan:
CI will probably pass once https://github.com/openshift/release/pull/39614 is merged.

PR #593, opened 2023-05-29
Adding an e2e job for the hub-spoke topology. (#385)

ybettan:
This workflow is using a single cluster which is both the hub and also the spoke and test the main flow of applying a `ManagedClusterModule` to the hub and validating that all the artifacts are created on the spoke cluster.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/572

### DIFF
--- a/ci/acm_subscription.yaml
+++ b/ci/acm_subscription.yaml
@@ -1,0 +1,20 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: acm-operator-group
+  namespace: advanced-cluster-management
+spec:
+  targetNamespaces:
+  - advanced-cluster-management
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: acm-operator-subscription
+  namespace: advanced-cluster-management
+spec:
+  channel: release-2.7
+  installPlanApproval: Automatic
+  name: advanced-cluster-management
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
--- a/ci/install-ci-hub/kustomization.yaml
+++ b/ci/install-ci-hub/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-kmm
+
+resources:
+  - ../../config/default-hub
--- a/ci/prow/e2e-hub-spoke-incluster-build
+++ b/ci/prow/e2e-hub-spoke-incluster-build
@@ -1,0 +1,67 @@
+#!/bin/bash
+#
+set -euxo pipefail
+
+readonly SKIP_MAKE_DEPLOY="${SKIP_MAKE_DEPLOY:-false}"
+readonly SKIP_ACM_INSTALLATION="${SKIP_ACM_INSTALLATION:-false}"
+
+export CLUSTER_NAME=local-cluster
+export OPERATOR_NAMESPACE=openshift-kmm
+export ACM_NAMESPACE=advanced-cluster-management
+
+
+if [ "$SKIP_ACM_INSTALLATION" == true ]; then
+  echo "Skipping ACM installation"
+else
+  echo "Install ACM..."
+  oc create namespace ${ACM_NAMESPACE}
+  oc apply -f ci/acm_subscription.yaml
+  timeout 1m bash -c '
+    until oc get crds -o json | jq -er ".items[].metadata.name | select(.? | match(\"multiclusterhubs\"))"; do
+      sleep 1;
+    done
+  '
+  oc apply -f ci/acm_multiclusterhub.yaml
+  timeout 15m bash -c '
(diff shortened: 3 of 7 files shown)


ybettan:
/hold
Didn't test it using the bundle (as we are going to do it CI) yet.

ci/acm_subscription.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: acm-operator-subscription
+  namespace: advanced-cluster-management
+spec:
+  channel: release-2.7
+  installPlanApproval: Automatic
+  name: advanced-cluster-management
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: advanced-cluster-management.v2.7.3
```
qbarrand, line 21: Is this field needed?

ci/prow/e2e-hub-spoke-incluster-build
```diff
@@ -0,0 +1,59 @@
...
+  oc apply -f ci/acm_multiclusterhub.yaml
+  acm_phase=$(oc -n advanced-cluster-management get multiclusterhub/multiclusterhub -o json | jq -r '.status.phase')
+  timeout 30m bash -c 'until [[ ${acm_phase} == "Running" ]]; do sleep 1; done'
+fi
+
+if [ "$SKIP_MAKE_DEPLOY" == true ]; then
+  echo "Skipping KMM deployment"
+else
+  echo "Deploy KMMO..."
+  export KUSTOMIZE_CONFIG_HUB_DEFAULT=ci/install-ci-hub
```
qbarrand, line 29:
```suggestion
  export KUSTOMIZE_CONFIG_HUB_DEFAULT=config/default-hub
```

ci/acm_subscription.yaml
ybettan: I am not sure, didn't try without it yet.
Assuming this is a dependency, don't we want to pin point its version instead of using the latest CSV?

ci/install-ci-hub/kustomization.yaml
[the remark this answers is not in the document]
ybettan: Because it is overriding the namespace of the hub to be installed in `openshift-kmm` instead of `openshift-kmm-hub` when deploying from makefile (this is not needed when deploying from bundle).

The reason I am overriding this namespace in the CI is because images built and pushed to the internal image registry are only accessible (by default) by apps deployed in the same namespace.

Another option will be to add a role binding to allow kmm to pull images from the kmm-hub namespace for example but my first try trying to use the `system:image-puller` role didn't succeed so I updated the namespace as it seemed simpler for the CI purposes. (this is also closes to what we are doing u/s).

ci/prow/e2e-hub-spoke-incluster-build
```diff
@@ -0,0 +1,59 @@
...
+
+
+if [ "$SKIP_ACM_INSTALLATION" == true ]; then
+  echo "Skipping ACM installation"
+else
+  echo "Install ACM..."
+  oc create namespace advanced-cluster-management
+  oc apply -f ci/acm_subscription.yaml
+  timeout 1m bash -c \
+      'until oc get crds -o json | jq -er ".items[].metadata.name | select(.? | match(\"multiclusterhubs\"))"; do sleep 1; done'
```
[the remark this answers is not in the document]
ybettan: I think `oc wait` is failing if the resource doesn't exist:
https://stackoverflow.com/questions/68226288/kubectl-wait-not-working-for-creation-of-resources

This is why we are using `until` in some places in the scripts.

ci/acm_subscription.yaml
qbarrand: I think we could work with the latest 2.7 release. I would even consider using the `stable` channel if it exists.

ci/prow/e2e-hub-spoke-incluster-build
```diff
@@ -0,0 +1,59 @@
...
+  oc apply -f ci/acm_multiclusterhub.yaml
+  acm_phase=$(oc -n advanced-cluster-management get multiclusterhub/multiclusterhub -o json | jq -r '.status.phase')
+  timeout 30m bash -c 'until [[ ${acm_phase} == "Running" ]]; do sleep 1; done'
```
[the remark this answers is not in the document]
ybettan: This is a very good questions! I will check it and fix it if this is the case.

ci/prow/e2e-hub-spoke-incluster-build
ybettan: `config/default-hub` will install in `openshift-kmm-hub`, `ci/install-ci-hub` will override it to be installed in `openshift-kmm` as mentioned in a previous comment.

ci/prow/e2e-hub-spoke-incluster-build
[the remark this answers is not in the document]
ybettan: It was indeed executed only once, I have updated it to be evaluated on each iteration.

qbarrand:
@ybettan can we unhold?

ybettan:
> @ybettan can we unhold?

Nope. It is not working yet.

PR #594, opened 2023-05-30
Adding OCP builds filtering by owner

yevgeny-shnaidman:
OCP builds are used during KMM sign and build flows. It can be created both by module reconcile flow and preflight flow. This PR adds additional filtering when looking for OCP builds based on module/kernel/type label to also filter by owner

### DIFF
--- a/internal/utils/ocpbuild/helper.go
+++ b/internal/utils/ocpbuild/helper.go
@@ -16,8 +16,8 @@ var ErrNoMatchingBuild = errors.New("no matching Build")
 //go:generate mockgen -source=helper.go -package=ocpbuild -destination=mock_helper.go
 
 type OCPBuildsHelper interface {
-	GetModuleOCPBuildByKernel(ctx context.Context, mld *api.ModuleLoaderData) (*buildv1.Build, error)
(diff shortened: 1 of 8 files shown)


internal/utils/ocpbuild/helper.go
```diff
@@ -33,7 +33,7 @@ func NewOCPBuildsHelper(client client.Client, buildType string) OCPBuildsHelper
 	}
 }
 
-func (o *ocpBuildsHelper) GetModuleOCPBuildByKernel(ctx context.Context, mld *api.ModuleLoaderData) (*buildv1.Build, error) {
+func (o *ocpBuildsHelper) GetModuleOCPBuildByKernel(ctx context.Context, mld *api.ModuleLoaderData, owner metav1.Object) (*buildv1.Build, error) {
 	buildList := buildv1.BuildList{}
 
 	opts := []client.ListOption{
```
qbarrand, line 39: Could we use a `fieldSelector` instead of filtering manually?

yevgeny-shnaidman: i don't think you can quire using Owner field, field selector is actually limited to which fields you can use

PR #605, opened 2023-06-05
Update skipper configuration (#422)

yevgeny-shnaidman:
1) removing operator-sdk, kustomize and mockgen, since makefile is
   using the once installed under the repo/bin directory
2) update the mockgen version


yevgeny-shnaidman:
Fixes [604](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/604

PR #646, opened 2023-06-23
Revert modprobe quotes and DaemonSet role labels

qbarrand:
Remove label and spec changes since 1.0 that prevented the DaemonSets from being reconciled.

/cc @yevgeny-shnaidman


bthurber (approved):
/lgtm

PR #647, opened 2023-06-25
Fixing DevicePlugin generation for ordered upgrade

yevgeny-shnaidman:
In case of ordered upgrade, we need to create a second DevicePlugin daemonset, since it will be managed by a different set of labels

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -383,7 +383,7 @@ func (mrh *moduleReconcilerHelper) handleDevicePlugin(ctx context.Context, mod *
 	if ds == nil {
 		logger.Info("creating new device plugin DS", "version", mod.Spec.ModuleLoader.Container.Version)
 		ds = &appsv1.DaemonSet{
-			ObjectMeta: metav1.ObjectMeta{Namespace: mod.Namespace, Name: mod.Name + "-device-plugin"},
+			ObjectMeta: metav1.ObjectMeta{Namespace: mod.Namespace, GenerateName: mod.Name + "-device-plugin-"},
 		}
 	}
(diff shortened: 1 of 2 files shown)


qbarrand:
This makes the device plugin restart on 1.0 to 1.1 upgrades.

yevgeny-shnaidman:
> This makes the device plugin restart on 1.0 to 1.1 upgrades.

The plugin won't restart , there will just be another Device Plugin Daemonset running.  We can add a patch that will remove the duplicate DaemonSet.  If we don't make the change to the name of the DevicePlugin DaemonSet, then we have a faulty ordered upgrade. WDYT?

yevgeny-shnaidman:
> > This makes the device plugin restart on 1.0 to 1.1 upgrades.
> 
> The plugin won't restart , there will just be another Device Plugin Daemonset running. We can add a patch that will remove the duplicate DaemonSet. If we don't make the change to the name of the DevicePlugin DaemonSet, then we have a faulty ordered upgrade. WDYT?

Thinking about it a little bit more:  after upgrade, in case device plugin DS was already running, we don't create a new one, and it will not support ordered upgrade (which is ok). In case of a new device plugin daemonset, that we need its name to be generated, so that we can support ordered upgrade, but it also means that that device plugin will be create only after upgrade. So we are good in both scenarious.

ybettan:
/gtm
/hold
@qbarrand unhold if you are happy with the change.

qbarrand:
lgtm. @yevgeny-shnaidman please unhold at will once you have checked the DaemonSet GC.

yevgeny-shnaidman:
> lgtm. @yevgeny-shnaidman please unhold at will once you have checked the DaemonSet GC.

I have verified that GarbageCollector is handling it. Please lgtm ( @ybettan had a typo in hist lgtm comment)

PR #657, opened 2023-07-10
MGMT-15181: Use ubi-minimal as the final base image

qbarrand:
Switch from ubi-micro that does not include libcrypto.
Unset CGO_ENABLED=0 to link against ubi's libraries instead of using Go's crypto.

/cc @yevgeny-shnaidman @mresvanis

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,9 +26,9 @@ ARG TARGET
 
 # Build
 RUN git config --global --add safe.directory ${PWD}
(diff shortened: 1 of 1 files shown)


ybettan:
Should we mention in the commit message that this is required to be FIPS compliant?

qbarrand:
@ybettan not sure I would do that. As is, KMM is released in ubi9 which is not FIPS compliant itself.

PR #667, opened 2023-07-23
NodeModulesConfig type refactoring (#481)

yevgeny-shnaidman:
1. Renaming NodeModulesState to NodeModulesConfig
2. adding ModuleConfig structure to NodeModuleSpec

this struct will hold the desired configuration of the module on the node (image, to be loaded or not, modprobe configuration etc'), and will be passed as is to the worker Pod as an input


yevgeny-shnaidman:
Fixes [665](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/665)

PR #671, opened 2023-07-26
introducing nmc package (#487)

yevgeny-shnaidman:
this package is used for creating, getting and manipulating the NodeModulesConfig objects. Currently the added function are the ones required by the module-nms reconcilaition, but will be expanded in the subsequent commits


yevgeny-shnaidman:
Fixes [670](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/670)

PR #673, opened 2023-07-27
Define version and commit through ldflags

qbarrand:
Do not use Go's `runtime/debug` to embed VCS information. Instead, set those values from external sources via ldflags.
- when building with `go build ./cmd/manager{,-hub}`, the value will be `undefined`
- when building with `make manager{,-hub}`, the `Makefile` will populate the version and git commit
- in CI, we fallback to `make manager` which should set the values
- on CPaaS, [$MR](https://gitlab.cee.redhat.com/rh-ecosystem-edge/kmm/-/merge_requests/233) adds the ldflags and uses CPaaS values to set the upstream commit.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -62,7 +62,12 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
-var scheme = runtime.NewScheme()
+var (
+	GitCommit = "undefined"
+	Version   = "undefined"
+
+	scheme = runtime.NewScheme()
+)
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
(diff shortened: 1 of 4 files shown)


cmd/manager/main.go
```diff
@@ -62,7 +62,12 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
-var scheme = runtime.NewScheme()
+var (
+	GitCommit = "undefined"
+	Version   = "undefined"
+
+	scheme = runtime.NewScheme()
+)
```
ybettan, line 70: Do we have to duplicate the var definition between here and manager-hub?

qbarrand: We could have those variables in a common package, but the ldflags would then look like `-X internal.cmd.Version=1.0.0`. Would you prefer that? IMO the variables should be related to the binary in an obvious way - I am concerned that this change would make things slightly more confusing.

ybettan: How about we set `main.Version` in manager and manger-hub to `internal.cmdutils.Version`? This way we remove the dup but doesn't change the interface in the Makefile?

qbarrand: So we still do `-X main.Version=1.0.0` in the ldflags, and then in the code that initializes `internal.cmdutils.Version` as well? I don't see the point of doing that. @yevgeny-shnaidman WDYT?

ybettan: > So we still do -X main.Version=1.0.0 in the ldflags

Yes.

I mean instead of having `Version = "undefined` in the manger, we set `Version = cmdutils.Version` and set `Version = "undefined"` in cmdutils.Version.

PR #688, opened 2023-08-02
Introducing Module-to-NMC reconciliation logic (#495)

yevgeny-shnaidman:
NMC package will update/create module's NMC entry
based on the mld configuration. Currently it is a standalone package with unit test. Later on it will be called by the module controller instead of HandleDriverContainer.
GarbageCollecto scenario is not handled by this PR: in case module does not contain mapping for the node's kernel version, but the NMC of the node does contain module version.

Decision flow:
1) based on node's kernel version, taints and labels, the module is
   either flag as should be running on the node or should not be running
   on the node
2) in case it should be running: the module config data of the module
   for that node's NMC is updated/created. In case NMC does not exists -
   it is creates first
3) in case it should not be running: if NMC does not exists, nothing to
   do,otherwise remove modules data from NMC spec


yevgeny-shnaidman:
Fixes [686](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/686)

qbarrand:
Bot didn't add the labels automatically so:
/ok-to-test
/approve
On my behalf:
/lgtm

PR #694, opened 2023-08-08
Introducing Module-NMC controller (#503)

yevgeny-shnaidman:
This PR introduces a new controller, that is responsible for reconciliation of NMC based on the data in Module and Node The following are the main points:
1) It watches for changes in Modules, Nodes (labels, becoming
   schedulable) and Jobs belonging owned by Modules (in order to
   reschedule once Build/Sign job are finished)
2) Per reconciled Module, go over all the nodes in the cluster
   and reconcile their NMCs based on the Module/Node state

Changes in the PR:
1) adding functions to filter package that are resposible for filtering
    the events (nodes. modules, jobs) and finding out which modules
    should be reconciled
2)  changing the NMC handler logic a little, adding verification for
    container image presence in case module should be enable.
3)  Module-NMC Reconciliation is being done both in case Kernel Mapping
    exists and in case it does not exists ( solves the garbage
    collection flow)

This PR does not include Module status updates based on NMC updates


yevgeny-shnaidman:
Fixes [692](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/692)

PR #696, opened 2023-08-08
In case NMC does not exists, Get should return NotFound error (#510)

yevgeny-shnaidman:


yevgeny-shnaidman:
Fixes [693](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/693)

PR #700, opened 2023-08-14
Module deleted scenario in Module-NMC controller (#518)

yevgeny-shnaidman:
When a Module is deleted, all its ModuleConfigs in all the NMCs should be removed. In order not to miss the delete event when the controller is dow, we do the following:
1) adding finalizer to the Module
2) whenever module is being deleted we remove module from all
   relevant NMCs, and if all the operations are successfull,
   we remove the finalizer also


yevgeny-shnaidman:
Fixes [699](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/699)

PR #713, opened 2023-08-24
Fixing panic in case worker pod has failed (#523)

yevgeny-shnaidman:
The status of the working pod that has completed with error does not contain the valid Config field ( it is set only on success). When processing NMC, the NMC controller tries to compare the Config of the status and spec, and in that case it fails, since status Config is nil.
Fix: in case status != nil, inProgress = false and Config is nil, just create a loader worker pod, without futher comparison to the spec


yevgeny-shnaidman:
Fixes [707](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/707)

PR #728, opened 2023-08-27
Adding Makefile target for worker image creation (#532)

yevgeny-shnaidman:
1) add target for worker image creation
2) add worker exe to .gitignore
3) update skipper configuration


yevgeny-shnaidman:
Fixes [716](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/716)

PR #731, opened 2023-08-27
Fixing permission for images' pull directory on worker pod (#522)

yevgeny-shnaidman:
Worker Pod container application tries to create directory for image pulling on /mnt/img, which fails due to the permission This PR add RunAsUser 0 to the security context


yevgeny-shnaidman:
Fixes [723](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/723)

PR #732, opened 2023-08-27
Reconfigure worker pod capabilities (#537)

yevgeny-shnaidman:
1) Removing privileged and adding SYS_MODULE capability
2) Worker application in worker pod will run as User 0 in order to
   be able to run modprobe


yevgeny-shnaidman:
Fixes [729](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/729)

PR #739, opened 2023-08-30
Adding Garbage Collection for NMC objects for deleted Nodes (#543)

yevgeny-shnaidman:
This implemented by setting Node as the owner of the NMC. In case Node is deleted, NMC will also be deleted


yevgeny-shnaidman:
Fixes [738](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/738)

PR #743, opened 2023-08-31
MGMT-15675: Mount OpenShift CAs into the worker Pod

qbarrand:
Create Cluster and Service CA ConfigMaps in each namespace containing at least one Module.
Mount those CAs into the worker Pod at `/etc/pki/tls/certs`.


cmd/manager/main.go
```diff
@@ -178,14 +181,16 @@ func main() {
 		filterAPI,
 		authFactory,
 		scheme,
+		operatorNamespace,
```
yevgeny-shnaidman, line 184: what do you say about making operatorNamespace a constant ("openshift-kmm")?  Why would a customer want to deploy it in a different one?

internal/controllers/module_nmc_reconciler.go
```diff
@@ -302,6 +321,7 @@ func (mnr *ModuleNMCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.
 		NewControllerManagedBy(mgr).
 		For(&kmmv1beta1.Module{}).
+		Owns(&v1.ConfigMap{}).
```
yevgeny-shnaidman, line 324: why do we need to watch ConfigMaps?

internal/ocp/ca/helper.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (h *helperImpl) getOrCreateClusterCAConfigMap(ctx context.Context, namespace string) (*v1.ConfigMap, error) {
+	cm, err := h.getConfigMap(ctx, namespace, clusterCAType)
+	if err != nil {
+		if !errors.Is(err, errConfigMapNotFound) {
+			return nil, fmt.Errorf("could not get the cluster CA ConfigMap: %v", err)
+		}
+
+		cm = &v1.ConfigMap{
```
yevgeny-shnaidman, line 115: why do we want to create an empty ConfigMap?

cmd/manager/main.go
qbarrand: I don’t think we can force the user to install into `openshift-KMM`, so I would leave this as is. It is populated automatically by the downwards API.

internal/controllers/module_nmc_reconciler.go
qbarrand: We own them, and we want to reconcile them in case they were altered or deleted.

internal/ocp/ca/helper.go
qbarrand: OpenShift controllers populate the contents based on the label or annotation.

internal/controllers/module_nmc_reconciler.go
yevgeny-shnaidman: But we want to reconcile them only in case something changed in Module/Node/NMC. In case ConfigMap has changed, what do we care? It won't affect NMC in anyway, no?

internal/ocp/ca/helper.go
yevgeny-shnaidman: So we can still get the scenario where the first time we run worker Pods, they may fail, since the ConfigMaps are not updated yet, right?

internal/controllers/module_nmc_reconciler.go
qbarrand: Moved to a dedicated controller.

internal/ocp/ca/helper.go
qbarrand: We can still hit that.

internal/controllers/module_nmc_reconciler_test.go
```diff
@@ -174,6 +174,7 @@ var _ = Describe("ModuleNMCReconciler_Reconcile", func() {
 		gomock.InOrder(
 			mockReconHelper.EXPECT().getRequestedModule(ctx, nsn).Return(&mod, nil),
 			mockReconHelper.EXPECT().setFinalizer(ctx, &mod).Return(nil),
+
```
yevgeny-shnaidman, line 177: We can remove that, there are no code changes to the file

PR #748, opened 2023-09-04
Changing the Module-NMC reconciliation  logic (#521)

yevgeny-shnaidman:
Previous logic:
get all nodes, and per node decide if Module should be running or not and update NMC accordingly
New logic:
1. get nodes by module selector.
2. get NMCs based on the label of the Module (in case module is/should be deployed on the node, a label will be set in NMC).
3. Per node, get the kernel mapping. 4  If kernel mapping exists, then module should be run on the node.
5. Otherwise, check if NMC exists for that node.
6. If it does, then module should be deleted from the node
7. After all the nodes has been process, look at the remaining NMCs: in all those node, module should be deleted

In order to support this flow, whenever NMC is configured to deploy the module, a label: beta.kmm.node.kubernetes.io/nmc.<namespace>.name is added to the NMC. whenever NMC is configured to undeploy the module, the label is removed


yevgeny-shnaidman:
Fixes [746](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/746)

PR #752, opened 2023-09-05
Use the default ServiceAccount when creating NMC

qbarrand:
Using default instead of an empty string allows us to attach all pull secrets configured for that account when setting up the mounts for the worker Pod.


internal/controllers/nmc_reconciler.go
```diff
@@ -431,14 +432,14 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 		if status == nil {
 			status = &kmmv1beta1.NodeModuleStatus{
 				ModuleItem: kmmv1beta1.ModuleItem{
-					ImageRepoSecret:    nil,
-					Name:               modName,
-					Namespace:          modNamespace,
-					ServiceAccountName: "",
+					Name:      modName,
```
yevgeny-shnaidman, line 435: we need to remember to fix it in the upstream also

qbarrand: https://github.com/kubernetes-sigs/kernel-module-management/pull/573

PR #763, opened 2023-09-20
Finalize Module resources

qbarrand:
Wait for kmods to be unloaded on all nodes to delete a Module. This is implemented through 2 labels:
- `beta.kmm.node.kubernetes.io/<ns>.<name>.module-configured`
- `beta.kmm.node.kubernetes.io/<ns>.<name>.module-in-use`

Both labels are set by the Module-NMC reconciler when an entry is added to the NMC spec.
The first label is removed by the Module-NMC reconciler when, for any reason, the NMC is not targeted by the Module anymore. The second label is removed by the NMC reconciler when the unloading worker pod is successful and the kmod is not present in the spec list anymore.
For the Module finalizer to be removed, no NMC must carry neither of these 2 labels.

/cc @yevgeny-shnaidman


qbarrand:
Fixes MGMT-15813

PR #770, opened 2023-09-21
Adding support for loading multiple kernel modules(softdep) (#549)

yevgeny-shnaidman:


yevgeny-shnaidman:
Fixes [766](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/766)

PR #771, opened 2023-09-21
Implementing ordered upgrade feature for V2 (#556)

yevgeny-shnaidman:
Current implementation involves creating a daemonset per module version, and setting ModuleLoader/DevicePlugin labels for the version value into the Daemosets selector fields. NodeLabelModuleVersion controller adds/removes appropriate labels from nodes based on customer input.

In V2, device plugin ordered upgrade implmentation will remain the same. For kernel modules, module-nmc reconciler will decide how to change NMCs based on the labels' changes of the nodes (provided by NodeLabelModuleVersion controler).

Module-NMC will use the following decicion-making table to update NCM: 1) if module is targeting the node, but kernel is not suitable - delete
   from NMC
2) if module is targeting the node, but the version is missing from
   Module (not an ordered upgrade) - add to NMC
3) if module is targeting the node, module loader version label is
   present on the node and its value equal to Module's version - add to
   NMC
4) if module is targeting the node, module loader version label is
   present on the node and its value is not equal to Module's version
   (meaning that the old version should be running) - no update to NMC.
5) if module is targeting the node, module loader version label is not
   present on the node, and Module's version is defined (meaning that
   currently no version should be running on the node) - delete from NMC

### DIFF
--- a/internal/controllers/module_nmc_reconciler_test.go
+++ b/internal/controllers/module_nmc_reconciler_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/module"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/nmc"
 	"github.com/rh-ecosystem-edge/kernel-module-management/internal/registry"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/utils"
 	"go.uber.org/mock/gomock"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


yevgeny-shnaidman:
Fixes [768](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/768)

internal/controllers/module_nmc_reconciler_test.go
```diff
@@ -508,26 +499,26 @@ var _ = Describe("prepareSchedulingData", func() {
...
-	targetedNodes := []v1.Node{node}
-	mld := api.ModuleLoaderData{KernelVersion: "some version"}
+	//targetedNodes := []v1.Node{node}
+	mld := api.ModuleLoaderData{KernelVersion: "some version", Name: moduleName, Namespace: moduleNamespace}
 
 	It("failed to determine mld", func() {
 		currentNMCs := sets.New[string](nodeName)
 		mockKernel.EXPECT().GetModuleLoaderDataForKernel(&mod, kernelVersion).Return(nil, fmt.Errorf("some error"))
 
+		fmt.Printf("YEV - in test, before call: <%s>\n", targetedNodes[0].Status.NodeInfo.KernelVersion)
```
qbarrand, line 521: Can we remove this?

internal/controllers/module_nmc_reconciler_test.go
```diff
@@ -508,26 +499,26 @@ var _ = Describe("prepareSchedulingData", func() {
...
 		mockHelper = nmc.NewMockHelper(ctrl)
 		mnrh = newModuleNMCReconcilerHelper(clnt, mockKernel, nil, mockHelper, nil, scheme)
+		node = v1.Node{
+			ObjectMeta: metav1.ObjectMeta{Name: nodeName},
+			Status: v1.NodeStatus{
+				NodeInfo: v1.NodeSystemInfo{KernelVersion: kernelVersion},
+			},
+		}
+		targetedNodes = []v1.Node{node}
+		fmt.Printf("YEV - kernel version <%s>\n", node.Status.NodeInfo.KernelVersion)
```
qbarrand, line 509: Can we remove this?

yevgeny-shnaidman: Next cherry-pick is going to remove

internal/controllers/module_nmc_reconciler_test.go
yevgeny-shnaidman: Next cherry-pick is going to remove

PR #774, opened 2023-09-21
Changing daemonsets GarbageCollector for v2 (#558)

yevgeny-shnaidman:
KMM V2 does not contain Daemonset for ModuleLoader anymore, only for DevicePlugin. This PR changes implementatio n of GarbageCollector function in daemonset package and the appropriate unit-tests. The changes for the GC function include:
1) no need to handle Module Loader daemonsets, including Module Version
   use case
2) no need verifying kernel versions of the nodes, which means changes
   to the GC function input parameters


yevgeny-shnaidman:
Fixes [769](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/769)

PR #775, opened 2023-09-21
Continuing adaption of ordered upgrade to KMM V2 (#563)

yevgeny-shnaidman:
Main points:
1) Renaming the ordered upgrade internal label from module-loader to
   worker-pod
2) simplifying labels reconciliation: no more need to check if
   module-loader pod exists, now we only need to check device-plugin
   pods
3) unit tests adaptation


yevgeny-shnaidman:
Fixes [772](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/772)

PR #781, opened 2023-09-21
Changing the flow of SyncStatus to accomodate Delete Pod errors (#566)

yevgeny-shnaidman:
Currently, in DeletePod is eexcuted per each Pod, and then NMC status is patched once for the the Modules. In case NMC status patch has failed, but the Pod was already deleted, we cannot reconcile NMC status in the next reconciliatio loop, since we don't know what do have the Pod. Since the status will contain InProgress=True, the module will never be reconciled.
In the new flow, all the Pods are deleted only after NMC was successfully patched. In case Pod deletion failed, the reconcile loop will try again, without progressing to NMC spec/status reconciliation. It will also allow for manual intervantion, in case Pod is stuck on some weird error


yevgeny-shnaidman:
Fixes [776](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/776)

PR #791, opened 2023-09-27
Removing DaemonSet role from the DevicePlugin labels (#580)

yevgeny-shnaidman:


yevgeny-shnaidman:
Fixes [790](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/790)

PR #795, opened 2023-09-27
Adjusting pod-node reconciler to v2 (#578)

yevgeny-shnaidman:
pod-node-reconciler must now handle only device-plugin pod in order not to interfere with the NMC reconciler, which handles node labeling based on worker pod


yevgeny-shnaidman:
Fixes [792](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/792)

PR #822, opened 2023-10-10
Fixing 2 bugs related to failure scenarios (#599)

yevgeny-shnaidman:
First bug:
If during module deletion scenario, UnloadKmod function returns and error for some reason, the GarbageCollectInUseLabels and UpdateNodeLabelsAndRecordEvents onw't be called. This in turn means, that if there is a module that was being loaded/unloaded in parallel, the actual action will happen (load/unload from kernel), but the label in NMC and node won't be updated.
Solution: always continue with GarbageCollectInUseLabels and UpdateNodeLabelsAndRecordEvents and return a collective error later

Second bug:
In case ProcessModuleSpec from some spec/status has failed, the status would not be deleted from the statusMap, which will cause it later to be treated as orphaned status and the UnloadKmod to be called. Solution: always delete status from statusMap, even in case of a failure

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -132,10 +132,10 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 				errs,
 				fmt.Errorf("error processing Module %s: %v", moduleNameKey, err),
 			)
-
-			continue
 		}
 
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
Fixes [821](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/821)

PR #828, opened 2023-10-13
Splitting Module reconcile into BuildSign and DevicePlugin reconciler…

yevgeny-shnaidman:
…s (#602)

Currently Module reconciler contains 2 logical flow that are completely unrelated to one another: Buid/Sign flow and DevicePligin flow The following are changes for this commit:
1) create reconciler responsible only for Build/Sign flow
2) create reconciler responsible only for DevicePlugin flow
3) delete Module reconciler
4) delete daemonset package, as it is no longer needed
5) unit tests


internal/controllers/build_sign_reconciler.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	return &mod, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *BuildSignReconciler) SetupWithManager(mgr ctrl.Manager, kernelLabel string) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kmmv1beta1.Module{}).
+		Owns(&buildv1.Build{}).
+		Owns(&batchv1.Job{}).
```
qbarrand, line 325: No need to own Jobs.

internal/controllers/build_sign_reconciler.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+}
+
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules,verbs=get;list;watch;update;patch
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list;delete;watch
+//+kubebuilder:rbac:groups="build.openshift.io",resources=builds,verbs=get;list;create;delete;watch;patch
+//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
```
qbarrand, line 72: No need for this line.

yevgeny-shnaidman: fixed

internal/controllers/build_sign_reconciler.go
yevgeny-shnaidman: fixed

PR #860, opened 2023-10-24
adapting Worker Pods status report in Module to v.2 (#619)

yevgeny-shnaidman:
This commits adds implementation of reporting worker pods status into the Module's status.
The following fields are reported:
1) number of targeted nodes ( based on selector field) 2) number of desired nodes that kernel module should be deployed (based
   on NMCs configured)
3) current number of the nodes that the kernel module is already
   deployed to ( based on NMC spec and status configs)


yevgeny-shnaidman:
Fixes [858](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/858)

PR #868, opened 2023-11-06
Add labs for KMM v2.0

enriquebelarte:
Updated labs to match with changes at KMM v2.0.

### DIFF
--- a/docs/mkdocs/labsv2.md
+++ b/docs/mkdocs/labsv2.md
@@ -1,0 +1,430 @@
+# KMM v2.0 Labs
+
+KMM Labs is a series of examples and use cases of Kernel Module Management Operator running on an Openshift cluster.
+
+
+## Installation
+
+If you want to learn more about KMM and how to install it, you can refer to the [documentation](https://docs.openshift.com/container-platform/4.13/hardware_enablement/kmm-kernel-module-management.html). The documentation provides detailed information about KMM and step-by-step instructions for the installation process.
+
+
+## Driver Toolkit
+
+We could use any image that includes the necessary libraries and dependencies for building a kernel module for our specific version. However, for ease, we can leverage Driver Toolkit (DTK), which is a container image used as a base image for building kernel modules.
+
+When using DTK, we should set which version of the Driver Toolkit [image](https://github.com/openshift/driver-toolkit#finding-the-driver-toolkit-image-url-in-the-payload) we are using to match the exact kernel version running on OpenShift nodes. In our case, we are using DTK with KMM and KMM will automatically determine what DTK digest should be used if the `DTK_AUTO` build argument is specified in the Dockerfile.
+
+We can get further info at DTK [repository](https://github.com/openshift/driver-toolkit#readme).
+
+
+## Building In-Cluster Modules from Sources
+
+In this example, we will demonstrate how to build the kernel module in our cluster using sources from a git repository. For this and next examples it is assumed that a namespace `labsv2` and a ServiceAccount `labsv2sa` allowed to use the privileged SCC have been previously created.
+
+The process involves several steps:
+
+1. Adding a ConfigMap: We will create a ConfigMap to store all the necessary information for the build process, including the image we are using (DTK in this case), the remote git repository containing the sources, and the compiling commands. It contains the `Dockerfile` used in the build process.
+
+2. Creating the Module object: Next, we'll create the Module object itself. It will specify the image to be built and used, a secret for possible authentication, ConfigMap for building process, and the nodes where the pods should be scheduled.
+
+Applying next file in an OpenShift cluster should build and load a kmm-ci-a module in the nodes labeled with a `worker` role:
+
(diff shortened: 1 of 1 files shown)


docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+We can get further info at DTK [repository](https://github.com/openshift/driver-toolkit#readme).
+
+
+## Building In-Cluster Modules from Sources
+
+In this example, we will demonstrate how to build the kernel module in our cluster using sources from a git repository. For this and next examples it is assumed that a namespace `labsv2` and a ServiceAccount `labsv2sa` allowed to use the privileged SCC have been previously created.
+
+The process involves several steps:
+
+1. Adding a ConfigMap: We will create a ConfigMap to store all the necessary information for the build process, including the image we are using (DTK in this case), the remote git repository containing the sources, and the compiling commands.
```
yevgeny-shnaidman, line 26: I would add that the ConfigMap contains the actual Dockerfile that will be used for build purposes, to make it more clear

docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+
+2. Creating the Module object: Next, we'll create the Module object itself. It will configure the registry for the resulting image, set up a secret for possible authentication, specify the previously created ConfigMap for use, and define the nodes where the pods should be scheduled.
```
yevgeny-shnaidman, line 28:
```suggestion
2. Creating the Module object: Next, we'll create the Module object itself. It will specify the image to be built and used, a secret for possible authentication, ConfigMap for building process, and the nodes where the pods should be scheduled.
```

docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+```
+
+## Device Plugins
+
+
+Device plugins play a crucial role in allocating and reporting specific hardware devices to the cluster. In KMM, you can load device plugin images as part of the Module object.
+
+For this example, we will utilize this plugin called [simple-device-plugin](https://github.com/yevgeny-shnaidman/simple-device-plugin) to simulate device plugins. This plugin provides a way to emulate device plugins and test their functionality within the cluster.
+
+It just loads some dummy devices settings in nodes as if they were reported by real hardware and then used by the moduleLoader.
```
yevgeny-shnaidman, line 239: this sentence is not clear to me: what component is loading dummy device settings on the node?

docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+  hugepages-2Mi              0 (0%)        0 (0%)
+  dummy/dummyDev             0             0
+  example.com/ptemplate      0             0
+  example.com/simple-device  0             0
+...
+```
+
+##  Seamless upgrade of kernel Module: Orderly process without reboot
+
+To minimize the impact on the workload running in the cluster, kernel module upgrades should be performed on a per-node basis.
```
yevgeny-shnaidman, line 293:
```suggestion
To minimize the impact on the workload running in the cluster, kernel module upgrades should be performed on a per-node basis in an order that can be defined/managed by user
```

docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+```
+kmm.node.kubernetes.io/version-module.<module-namespace>.<module-name>=$moduleVersion
+```
+
+In our case we'll use this example:
+```bash
+oc label node my-node-1 kmm.node.kubernetes.io/version-module.labsv2.kmm-ci-a=0.9
+```
+
+Then we can use a new build based on previous build example by just adapting the ModuleLoader (adding `version` and specific `containerImage` for that version) and reusing existing Configmap for build. Make sure you have deleted any Module object left from past examples and apply following Module:
```
yevgeny-shnaidman, line 306: I think we should mention that  "ordered upgrade" works only in case "version field was defined in the Module from the very-begining: it won't work on an already existing Module without version field, which will be added during the Module edit

docs/mkdocs/labsv2.md
```diff
@@ -0,0 +1,425 @@
...
+   (In our example 1.0 for the version and `kmm-kmod-${KERNEL_FULL_VERSION}-v1.0` for the containerImage):
+```
+oc edit module kmm-ci-a
+```
+6. The new module version is loaded.
+
+
+## Loading soft dependencies
+
+Kernel modules sometimes have dependencies on others modules so those dependencies have to be loaded in advance. In the following example `kmm-ci-a` depends on `kmm-ci-b` so we will set `modulesLoadingOrder` and then the list with the `moduleName` as the first entry followed by all of its dependencies. In our case it is just `kmm-ci-b` but it could be a longer list with multiple dependency modules.
```
yevgeny-shnaidman, line 360: we should add that soft dependency should be used when there is a logical dependency between modules,but no  symbols' dependency (otherwise modprobe  would have done the work automatically)

PR #874, opened 2023-11-08
Move to Go 1.20

### LINKED ISSUE #712: Move to `errors.Join` when we move to Go 1.20
Upstream uses [`errors.Join`](https://pkg.go.dev/errors#Join) to group errors into return values, but that is only available in Go 1.20+.
When we move to that version, replace usages of [multierror](https://github.com/hashicorp/go-multierror) with `errors.Join`.

qbarrand:
Use `errors.Join` instead of `multierror`.

Fixes #712
Fixes #873

/cc @yevgeny-shnaidman


qbarrand:
Requires https://github.com/openshift/release/pull/45422 to be merged first.

yevgeny-shnaidman:
lint is failing because of fmt

Issue #920, opened 2023-12-14
KMM Day 1 services should be uniquely named


pcolledg-amd:
User may deploy multiple modules to nodes. Currently the user would need to manually rename `replace-kernel-module.service` & `pull-kernel-module-image.service` for each unique module. This works, but standardising on a renaming method would aid documentation and avoid accidental overwriting. Renaming could either 

* be a prefix/suffix (eg. `mod_a-replace-kernel-module.service`), or 
* use systemd Service Templates (eg. `replace-kernel-module@mod_a.service`) with image details specified in a templated file elsewhere (eg. `EnvironmentFile=.../%i.env`), which would make the .service files generic not templated.

On a related note, in my opinion, `/usr/local/bin/replace-kernel-module.sh` & `/usr/local/bin/pull-kernel-module-image.sh` files would be better distributed as generic not templated, configured entirely within the module-specific .service files by _runtime_ env vars (ie. systemd `Environment=`) and/or script command line args.

PR #939, opened 2023-12-21
Day1 KMM solution should support V2 with scratch images

yevgeny-shnaidman:
Currently day1 solution relies on the fact that images kernel module images used are DTK-based image, and therefore contain modprobe command. Since KMM V2 we now support basic scratch images. This PR adds support for scratch images by running worker image with supplied config


pkg/mcproducer/templates/replace-kernel-module.gotmpl
```diff
@@ -2,14 +2,14 @@
 echo "before checking podman images"
 if podman image exists {{.Image}}; then
     echo "Image {{.Image}} found in the local registry, removing in-tree kernel module"
-    podman run --privileged --entrypoint modprobe {{.Image}} -rd /opt {{.KernelModule}}
+    podman run --user=root --privileged -v /etc/kmm-worker-day1/config.yaml:/etc/kmm-worker/config.yaml {{.Image}} kmod unload /etc/kmm-worker/config.yaml
```
ybettan, line 5: What is the benefits of using `kmod` rather than `modprobe`?

yevgeny-shnaidman: this code actually runs worker pod, and kmod is part d the arguments to the worker app in the pod. modporbe will be called by the worker pod

ybettan: Ohh OK. This is confusing since there is a `kmod` binary loading kernel modules in Linux.

qbarrand:
/hold
Until the worker pod can load from a tarfile instead of pulling an image.

pkg/mcproducer/mcproducer.go
```diff
@@ -8,6 +8,11 @@ import (
 	"text/template"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	//utilrand "k8s.io/apimachinery/pkg/util/rand"
```
qbarrand, line 11: Please remove this.

pkg/mcproducer/mcproducer.go
```diff
@@ -75,14 +91,11 @@ func executeIntoBase64(tmpl *template.Template, params map[string]any) (string,
...
-	if tagErr == nil {
-		return nil
+func getLocalFileName(containerImage string) (string, error) {
+	_, err := name.ParseReference(containerImage)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse container image %s name: %v", containerImage, err)
 	}
-	return fmt.Errorf("invalid image %s, input should be either in tag or digest format. Digest error %v, Tag error %v", image, digestErr, tagErr)
+
+	return "/var/lib/image_file_day1.tar", nil
```
qbarrand, line 100: This restricts the day1 solution to one image only. Should we make that path unique? See #920.

pkg/mcproducer/mcproducer_test.go
```diff
@@ -35,6 +36,10 @@ var _ = Describe("ProduceMachineConfig", func() {
 		Expect(err).ToNot(HaveOccurred())
 		expectedRes, err := os.ReadFile("testdata/machineconfig-test.yaml")
 		Expect(err).ToNot(HaveOccurred())
+		if string(expectedRes) != res {
+			fmt.Printf("<%s>\n", res)
+			fmt.Printf("<%s>\n", expectedRes)
+		}
```
qbarrand, line 42: This looks like debugging to me 🙂

pkg/mcproducer/templates/pull-image.gotmpl
```diff
@@ -1,13 +1,22 @@
 #!/bin/bash
```
qbarrand, line 1: I believe it is systemd calling those scripts, right? If yes, I think it would be cleaner if the variables were either environment variables or arguments configured in the unit file.

pkg/mcproducer/mcproducer.go
yevgeny-shnaidman: I think that a better solution would be to use soft dependencies: saves loading different MCOs, creating new additional services and pulling different images

pkg/mcproducer/mcproducer.go
yevgeny-shnaidman: fixed

pkg/mcproducer/mcproducer_test.go
yevgeny-shnaidman: It was done intentionally: in case test fails, it would be easier to debug seeing the whole ouput of both MCO

pkg/mcproducer/templates/pull-image.gotmpl
yevgeny-shnaidman: Not sure i agree: those are one-shot services. We don not expect to configure/change them once they have been run. In addition having all the parameters defined in the scripts allows for easier testing of scripts as standalone

qbarrand:
MGMT-16281

PR #949, opened 2024-01-02
Fixing DevicePlugin upgrade from v1.x to v2.x (#679)

yevgeny-shnaidman:
In v1.x we had 2 type of Daemonsets: DevicePlugin and ModuleLoader, which could be distingished by the DaemonsetRole label In v2.x, there only 1 type of daemonset, DevicePlugin, so the DaemonsetRole labels was removed.
During DevicePlugin reconciliation, reconcile get all the device plugin Ds based only on module name label, which means that in case there is a v1.1 ModuleLoader DS running it will also be chosen. Once reconciler tries to reconcile ModuleLoader DS, it will fail
This PR, updates the function that chooses the DevicePlugin DS, by removing old ModuleLoader Ds based on Role label


qbarrand:
MGMT-16246

PR #957, opened 2024-01-10
Adding an enhancement for integrating day0 kmods with KMM.

ybettan:
---

/cc @yevgeny-shnaidman @qbarrand @mresvanis

### DIFF
--- a/docs/enhancements/0002_day0-integration.md
+++ b/docs/enhancements/0002_day0-integration.md
@@ -202,8 +202,9 @@ with the new image.
 ### Firmware
 
 No much to do here, the user can put his firmware files directly in
-`/var/lib/firmware` on the nodes in the custom container image so the firmware
-feature of KMM is a bit useless here.
+`/lib/firmware` on the nodes in the custom container image or set
+`firmware_class.path` in their image's `sysctl.conf` so the firmware feature of
+KMM is a bit useless here.
 
 ### Preflight
 
--- a/docs/enhancements/0002_day0-integration.md
+++ b/docs/enhancements/0002_day0-integration.md
@@ -62,31 +62,9 @@ KMM will watch for `MachineConfig`s in the cluster containing the
 `machineconfiguration.openshift.io/kmm-managed: "true"` label and create a new
 `ModuleDay0` CR in the cluster for each one of them.
 We will have a different controller for this new CRD in KMM.
-**There should be no more than 1 `ModuleDay0` targeting a `MachineConfigPool`
-(masters/workers).
-If multiple kmods are required they should all be backed in the same image**
-
-Since MCO can only apply `MachineConfig`s changes to an entire `MachineConfigPool`,
-we must require that a `ModuleDay0` will target a `MachineConfigPool` as well.
-MCO ties between the `machineconfiguration.openshift.io/role: worker` label on
-the `MachineConfig` and the `node-role.kubernetes.io/worker: ""` label on the
-node.
-
-A customer can always create his own `MachineConfigPool` and add only 1 worker
-to it but I don't think this should be one of our primary concerns ATM for the
-following reasons
-  * assisted-installer will generate the same ISO for all nodes in the cluster
-    anyway so there is no benefit of having finer grained node selector for
-    `ModuleDay0` when assisted doesn't support such resolution on its own.
-  * if a user wishes to have only some worker nodes modified, he will need to
-    create is own `MachineConfigPool` post cluster installation and then update
-    the `machineconfiguration.openshift.io/role: worker` label of the
-    `MachineConfig` to target the new `MachineConfigPool`.
-    This will make MCO role back the nodes that shouldn't be updated to the
-    payload OS.
-    Using layering for only some of the workers will result in a cluster upgrade
-    that only upgrade the nodes that aren't layered which may cause some
-    inconsistency and confusion
+**There should be no more than 1 `ModuleDay0` per `MachineConfig`.
+If multiple kmods are required they should all be backed in the same image and
+used in the same `MachineConfig` by the user**
 
 After the `ModuleDay0` was create by KMM, all KMM's features will be used by the
 official KMM's API which is the `ModuleDay0` objects in the cluster.
@@ -106,8 +84,7 @@ spec:
   osMapping:
     literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
     containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
-  machineConfigPoolSelector:
-    node-role.kubernetes.io/worker: ""
+  machineConfigName: 99-ybettan-external-image
 ```
 
 * The `containerImage` here must be a full RHCOS image built by correct tools
@@ -149,8 +126,7 @@ spec:
         - name: build-secret
       dockerfileConfigMap:
         name: kmm-kmod-dockerfile
-  machineConfigPoolSelector:
-    node-role.kubernetes.io/worker: ""
+  machineConfigName: 99-ybettan-external-image
 ```
 
 The build will trigger if all the following conditions are met
@@ -212,8 +188,7 @@ spec:
         name: kmm-kmod-signing-key
       filesToSign:
         - /opt/lib/modules/${KERNEL_FULL_VERSION}/kmm_ci_a.ko
-  machineConfigPoolSelector:
-    node-role.kubernetes.io/worker: ""
+  machineConfigName: 99-ybettan-external-image
 ```
 
 The sign will trigger if all the following conditions are met
@@ -273,8 +248,7 @@ spec:
     literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
     containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
     clusterUpgradeTriggerNodeUpgrade: true
-  machineConfigPoolSelector:
-    node-role.kubernetes.io/worker: ""
+  machineConfigName: 99-ybettan-external-image
 ```
 
 ### Hub&Spoke
--- a/docs/enhancements/0002_day0-integration.md
+++ b/docs/enhancements/0002_day0-integration.md
@@ -1,0 +1,282 @@
+# Day-0 integration with KMM
+
+### Motivation
+
+KMM is managing mostly day2 kmods in OCP. It also offers a library for generating
+systemd files for helping customers to deploy their day-1 kmod.
+
+What KMM doesn't offer today to customers is the deployment and management of
+day-0 kmods.
+
+Day-0 kmod are complex as they need to be present in the system at boot time,
+even before OCP is fully installed and certainly before KMM is deployed to the
+cluster, therefore, there is not much that KMM can do to help deploy those kmods but
+it can manage their lifecycle once a customer has installed a cluster with his
+day-0 kmods in it - this is what this enhancement is about.
+
+### Day-0 cluster background
+
+A customer wishing to deploy OCP with custom nodes will need to generate 2
+artifacts using [image-builder](https://www.redhat.com/en/topics/linux/what-is-an-image-builder)
+  * A custom ISO for the nodes
+  * A container image encapsulating the ostree-commit that is deployed in the ISO
+
+The ISO will be used by [assisted-installer](https://docs.openshift.com/container-platform/4.14/installing/installing_on_prem_assisted/installing-on-prem-assisted.html)
+to install the cluster and the container image will be used to notify
+[MCO](https://docs.openshift.com/container-platform/4.14/post_installation_configuration/machine-configuration-tasks.html#understanding-the-machine-config-operator)
+that the image on the nodes is a custom image.
+
+When a customers has installed a "day-0 cluster" as mentioned above, we can
+assume that there is a `MachineConfig` in the cluster that looks something like
+```
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 99-ybettan-external-image
+spec:
+  osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0
+```
+The important piece in it is `osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0`
+which tells MCO that the ostree-commit that is supposed to be running on
+the node is encapsulated in `quay.io/ybettan/rhcos:414.92.202312311229-0`.
+
+If MCO finds out that the OS on the node is different than the image specified
+in `osImageURL` then it will override the node OS based on the `MachineConfig`
+and reboot - we are going to leverage this mechanism for the integration with
+KMM.
+
+In a day-0-cluster scenario, the user has applied this `MachineConfig` to tell
+MCO he has modified the OS on the nodes and that it shoudn't override it with
+the RHCOS image from the payload.
+
+### Setting up the integration
+
+A user that wishes to hand over the management of his day-0 kmod to KMM will
+first need to make sure the `MachineConfig` can be reconize by KMM, therefore,
+he would add a `machineconfiguration.openshift.io/kmm-managed: "true"` label to
+it.
+
+KMM will watch for `MachineConfig`s in the cluster containing the
+`machineconfiguration.openshift.io/kmm-managed: "true"` label and create a new
+`ModuleDay0` CR in the cluster for each one of them.
+We will have a different controller for this new CRD in KMM.
+**There should be no more than 1 `ModuleDay0` targeting a `MachineConfigPool`
+(masters/workers).
+If multiple kmods are required they should all be backed in the same image**
+
+Since MCO can only apply `MachineConfig`s changes to an entire `MachineConfigPool`,
+we must require that a `ModuleDay0` will target a `MachineConfigPool` as well.
+MCO ties between the `machineconfiguration.openshift.io/role: worker` label on
+the `MachineConfig` and the `node-role.kubernetes.io/worker: ""` label on the
+node.
+
+A customer can always create his own `MachineConfigPool` and add only 1 worker
+to it but I don't think this should be one of our primary concerns ATM for the
+following reasons
+  * assisted-installer will generate the same ISO for all nodes in the cluster
+    anyway so there is no benefit of having finer grained node selector for
+    `ModuleDay0` when assisted doesn't support such resolution on its own.
+  * if a user wishes to have only some worker nodes modified, he will need to
+    create is own `MachineConfigPool` post cluster installation and then update
+    the `machineconfiguration.openshift.io/role: worker` label of the
+    `MachineConfig` to target the new `MachineConfigPool`.
+    This will make MCO role back the nodes that shouldn't be updated to the
+    payload OS.
+    Using layering for only some of the workers will result in a cluster upgrade
+    that only upgrade the nodes that aren't layered which may cause some
+    inconsistency and confusion
+
+After the `ModuleDay0` was create by KMM, all KMM's features will be used by the
+official KMM's API which is the `ModuleDay0` objects in the cluster.
+
+In case a user will wish to stop the sync between MCO and KMM, he will need to
+remove the `machineconfiguration.openshift.io/kmm-managed: "true"` label from
+the `MachineConfig`; KMM will then delete the corresponding `ModuleDay0`.
+Once done, if the user delete the `MachineConfig` in the cluster then MCO will
+roll-back to the RHCOS image from the payload (without the kmods).
+
+### The `ModuleDay0` CRD
+
+```
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: ModuleDay0
+spec:
+  osMapping:
+    literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
+    containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
+  machineConfigPoolSelector:
+    node-role.kubernetes.io/worker: ""
+```
+
+* The `containerImage` here must be a full RHCOS image built by correct tools
+  or a container image based on an RHCOS container image and not an image based
+  on UBI.
+
+### Modifying day-0 kmods
+
+In day2, we have `kernelMappings` which defines what image should be use for
+each kernel (or future kernel) in the cluster. Once KMM notice a new kernel on
+some nodes, it will start acting on the relevant `kernelMapping`, for example,
+it will load the new kmod on the node.
+
+In day0 modules, we will take a pro-active approach - we will modify the nodes
+instead of waiting for them to get updated by an external process.
+We will have a single `osMapping` which its output artifcats should exist
+**before** we modify the node.
+
+If the `osMapping` in the `ModuleDay0` is different than the
+`node.status.nodeInfo.osImage` then we override `mc.osImageURL` with the image
+from `moduleday0.spec.osMapping.containerimage`.
+
+MCO will then override the node with the new content and potentialy reboot the
+node.
+
+### Building
+
+If a customer wishes KMM to re-build his kmods, they should add a `build` section
+in the `ModuleDay0` as follow
+```
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: ModuleDay0
+spec:
+  osMapping:
+    literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
+    containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
+    build:
+      secrets:
+        - name: build-secret
+      dockerfileConfigMap:
+        name: kmm-kmod-dockerfile
+  machineConfigPoolSelector:
+    node-role.kubernetes.io/worker: ""
+```
+
+The build will trigger if all the following conditions are met
+  * `quay.io/ybettan/rhcos:414.92.202312311229-0` is different from `spec.osImageURL` in the `MachineConfig`;
+  * `quay.io/ybettan/rhcos:414.92.202312311229-0` doesn't exist
+  * A `build` section exists in the `osMapping`
+
+Once built, KMM will modify the `mc.osImageURL` with the new image.
+
+The `ConfigMap` should be in the following format
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kmm-kmod-dockerfile
+data:
+  dockerfile: |
+
+    ARG DTK_AUTO
+    ARG OS_AUTO
+
+    FROM ${DTK_AUTO} as builder
+
+    ARG KERNEL_VERSION
+    ...
+    RUN KERNEL_SRC_DIR=/lib/modules/${KERNEL_VERSION}/build make all
+
+    FROM ${OS_AUTO}
+
+    ARG KERNEL_VERSION
+
+    COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_a.ko /opt/lib/modules/${KERNEL_VERSION}/
+    COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_b.ko /opt/lib/modules/${KERNEL_VERSION}/
+    RUN depmod -b /opt ${KERNEL_VERSION}
+```
+
+`OS_AUTO` will fetch the `rhel-coreos` image from the release-payload of the
+cluster.
+The DTK `ImageStream` on the cluster will contain the DTK image that fits the
+release-payload and not the running OS on the nodes (which may have a different
+kernel), therefore, `DTK_AUTO` should only be used if `OS_AUTO` is used.
+
+### Signing
+
+If a customer wishes KMM to re-build his kmods, they should add a `build` section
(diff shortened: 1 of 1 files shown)


ybettan:
Here is a link describing the full process on how to build the first custom ISO and container image to build the cluster using assisted-installer: https://github.com/ybettan/image-composer/blob/main/CUSTOMIZE_RHCOS.md

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+KMM is managing mostly day2 kmods in OCP. It also offers a library for generating
+systemd files for helping customers to deploy their day-1 kmod.
+
+What KMM doesn't offers today to customers is the deployment and management of
+day-0 kmods.
+
+Day-0 kmod are complex as they need to be present in the system at boot time,
+even before OCP is fully installed and certaily before KMM is deployed to the
+cluster, therefore, there is no much KMM can do to help deploy those kmods but
```
qbarrand, line 13:
```suggestion
cluster, therefore, there is not much that KMM can do to help deploy those kmods but
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+metadata:
+  creationTimestamp: "2023-12-31T14:03:08Z"
+  generation: 1
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 99-ybettan-external-image
+  resourceVersion: "1758"
+  uid: d1f07169-4cb5-4b85-93b3-3cd80530f028
+spec:
+  osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0
```
qbarrand, line 43:
```suggestion
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: worker
  name: 99-ybettan-external-image
spec:
  osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+```
+The important piece in it is `osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0`
+which tells MCO what that the ostree-commit that is supposed to be running on
+the node is encapsulated in `quay.io/ybettan/rhcos:414.92.202312311229-0`.
```
qbarrand, line 47:
```suggestion
The important piece in it is `osImageURL: quay.io/ybettan/rhcos:414.92.202312311229-0`
which tells MCO that the ostree-commit that is supposed to be running on
the node is encapsulated in `quay.io/ybettan/rhcos:414.92.202312311229-0`.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+If MCO, finds out that the OS on the node is different than the image specified
+in `osImageURL` then MCO will override the node OS based on the `MachineConfig`
+and reboot - we are going to leverage this mechanisem for the integration with
+KMM.
```
qbarrand, line 52:
```suggestion
If MCO finds out that the OS on the node is different than the image specified
in `osImageURL` then it will override the node OS based on the `MachineConfig`
and reboot - we are going to leverage this mechanism for the integration with
KMM.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+he would add a `machineconfiguration.openshift.io/kmm-managed: "true"` label to
+it.
+
+KMM will watch for `MachineConfig`s in the cluster containing the
+`machineconfiguration.openshift.io/kmm-managed: "true"` label and create a new
+`ModuleDay0` CR in the cluster for each one of them.
+We will have a different controller for this new CRD in KMM.
+**There should be no more than 1 `ModuleDay0` targeting a `MachineConfigPool`
+(masters/workers).
+If multiple kmods are required they should all be backed in the same image**
```
qbarrand, line 71: What happens today if two `MachineConfig` objects setting different `osImageURL` values target the same `MachineConfigPool`?

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+Since MCO can only apply `MachineConfig`s changes to an entire `MachineConfigPool`,
+we must require that a `ModuleDay0` will target a `MachineConfigPool` as well.
+MCO ties between the `machineconfiguration.openshift.io/role: worker` label on
+the `MachineConfig` and the `node-role.kubernetes.io/worker: ""` label on the
+node.
+
+A customer can always create his own `MachineConfigPool` and add only 1 worker
+to it but I don't think this should be one of our primary concerns ATM for the
+following reasons
```
qbarrand, line 81: Let's avoid first-person style in this doc.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+    `ModuleDay0` when assisted doesn't support such resolution on its own.
+  * if a user wishes to have only some worker nodes modified, he will need to
+    create is own `MachineConfigPool` post cluster installation and then update
+    the `machineconfiguration.openshift.io/role: worker` label of the
+    `MachineConfig` to target the new `MachineConfigPool`.
+    This will make MCO role back the nodes that shouldn't be updated to the
+    payload OS.
+    Using layering for only some of the workers will result in a cluster upgrade
+    that only upgrade the nodes that aren't layered which may cause some
+    inconsistency and confusion
```
qbarrand, line 93: Can assisted use the custom ISO for some nodes only?

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+After the `ModuleDay0` was create by KMM, all KMM's features will be used by the
+official KMM's API which is the `ModuleDay0` objects in the cluster.
+
+In case a user will wish to stop the sync between MCO and KMM, he will need to
+remove that label from the `MachineConfig` and KMM will delete its `ModuleDay0`.
```
qbarrand, line 99:
```suggestion
remove the `machineconfiguration.openshift.io/kmm-managed: "true"` label from the `MachineConfig`; KMM will then delete the corresponding `ModuleDay0`.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+* The `containerImage` here must be a full RHCOS image built by correct tools
+  or a container image based on an RHCOS container image and not an image based
+  on UBI.
+
+### Modifying day-0 kmods
+
+In day2, we have `kernelMappings` which defines what image should be use for
+each kernel (or future kernel) in the cluster. Once KMM notice a new kernel on
+some nodes, he will start acting on the relevant `kernelMapping`, for example,
+he will load the new kmod on the node.
```
qbarrand, line 125:
```suggestion
some nodes, it will start acting on the relevant `kernelMapping`, for example,
it will load the new kmod on the node.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+In day0 modules, we will take a pro-active approach - we will modify the nodes
+instead of waiting for them to get updated by an external process.
+The reason for that will be clear later on in this doc.
```
qbarrand, line 129:
```suggestion
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+If the `osMapping` in the `ModuleDay0` is different than the
+`node.status.nodeInfo.osImage` then we override `mc.osImageURL` with the image
+from `moduleday0.spec.osMapping.containerimage`.
+
+MCO will then override the node with the new content and potentialy reboot the
+node.
+
+### Building
+
+If a customer wish us to re-build his kmods, he should add a `build` section
```
qbarrand, line 142:
```suggestion
If a customer wishes to re-build his kmods, they should add a `build` section
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+```
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: ModuleDay0
+spec:
+  osMapping:
+    literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
+    containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
+    build:
+      secrets:
+        - name: build-secret
```
qbarrand, line 153: This will require a namespace as well as `ModuleDay0` doesn't seem to be namespaced.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+Once done, if the user delete the `MachineConfig` in the cluster then MCO will
+roll-back to the RHCOS image from the payload (without the kmods).
+
+### The `ModuleDay0` CRD
+
+```
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: ModuleDay0
+spec:
+  osMapping:
```
qbarrand, line 109: Do we need this `osMapping` if the `containerImage` is to be set in all nodes of the MCP?

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+      secrets:
+        - name: build-secret
+      dockerfileConfigMap:
+        name: kmm-kmod-dockerfile
+  machineConfigPoolSelector:
+    node-role.kubernetes.io/worker: ""
+```
+
+The build will trigger if all the following conditions are met
+  * `quay.io/ybettan/rhcos:414.92.202312311229-0` is different than `mc.osImageURL`
```
qbarrand, line 161:
```suggestion
  * `quay.io/ybettan/rhcos:414.92.202312311229-0` is different from `spec.osImageURL` in the `MachineConfig`;
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+      keySecret:
+        name: kmm-kmod-signing-key
+      filesToSign:
+        - /opt/lib/modules/${KERNEL_FULL_VERSION}/kmm_ci_a.ko
+  machineConfigPoolSelector:
+    node-role.kubernetes.io/worker: ""
+```
+
+The sign will trigger if all the following conditions are met
+  * `quay.io/ybettan/rhcos:414.92.202312311229-0` is different than `mc.osImageURL`
```
qbarrand, line 224:
```suggestion
  * `quay.io/ybettan/rhcos:414.92.202312311229-0` is different from `spec.osImageURL` in the `MachineConfig`;
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+  * `quay.io/ybettan/rhcos:414.92.202312311229-0-unsigned` exist (or a `build` section exists)
+  * A `sign` section exists in the `osMapping`
+
+Once signed and push, KMM will modify the `mc.osImageURL` with the new image.
```
qbarrand, line 228:
```suggestion
Once signed and pushed, KMM will modify the `MachineConfig`'s `spec.osImageURL` with the new image.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Firmware
+
+No much to do here, the user can put his firmware files directly in
+`/var/lib/firmware` on the nodes in the custom container image so the firmware
+feature of KMM is a bit useless here.
```
qbarrand, line 234: `/var/lib/firmware` is not a default lookup path for the kernel. Users should either:
- put the firmware files in `/lib/firmware`, checking carefully that they don't overwrite anything, or;
- set `firmware_class.path` in their image's `sysctl.conf`.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+When a `MachineConfig` uses `osImageURL`, it is called
+[image-layering](https://docs.openshift.com/container-platform/4.14/post_installation_configuration/coreos-layering.html).
+
+When image-layering is used, the OCP layer is being "detached" from the underlying
+OS running on the nodes. It means that when a cluster upgrade is triggered, for
+example from `4.14` to `4.15`, the OS on the nodes remain to be the same and the
+user only gets the new OCP features without the new RHCOS features.
+
+If a user has layered the images on the cluster, it is his responsibility to
+upgrade the nodes when he wishes.
```
qbarrand, line 255:
```suggestion
If a user has layered the images on the cluster, it is their responsibility to
upgrade the nodes when they wish.
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+In the day-0 case, KMM can enhance the UX and make the customer feel like a
+cluster upgrade is still upgrading the nodes.
+
+The way we can achieve that for new OCP y-streams as follow.
```
qbarrand, line 260:
```suggestion
The way we can achieve that for new OCP y-streams is as follows:
```

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+
+We check `clusterversion.status.desired.version` and `node.status.nodeInfo.osImage`
+and see if the y-stream in both versions differ.
+
+If they does, we can rebuild the container image (if the user has supplied the
```
qbarrand, line 265:
```suggestion
If they do, we can rebuild the container image (if the user has supplied the
```

yevgeny-shnaidman:
General remark: I still think that KMM  should not be the one responsible for updating the MachineConfig

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+(masters/workers).
+If multiple kmods are required they should all be backed in the same image**
+
+Since MCO can only apply `MachineConfig`s changes to an entire `MachineConfigPool`,
+we must require that a `ModuleDay0` will target a `MachineConfigPool` as well.
+MCO ties between the `machineconfiguration.openshift.io/role: worker` label on
+the `MachineConfig` and the `node-role.kubernetes.io/worker: ""` label on the
+node.
+
+A customer can always create his own `MachineConfigPool` and add only 1 worker
```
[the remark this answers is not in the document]
ybettan: > why do we care about MachineConfigPool?
The reason I mentioned `MachineConfigPool` explanation is to justify the need of having a single `ModuleDay0` per `MachineConfigPool`.

docs/enhancements/0002_day0-integration.md
ybettan: Nope. This is a good point. I will modify it in an additional commit so it is easier to review (will squash before merge).

docs/enhancements/0002_day0-integration.md
ybettan: It can but not in the way that assisted was designed to.
It will make more mess than order.
Should be simpler to just install the driver on all workers even if the node doesn't have the HW.
In addition to that, if an admin wish to add the HW later on then the driver is already in place.

If a user really want to be as precise as possible he can always follow:
1. Make all worker nodes custom
2. Once the cluster is installed, the user create a `MachineConfigPool` and add the worker he wishes to customize to it.
3. He modifies the `MachineConfig` to target the new pool he has created.
4. MCO reboot all other worker nodes to the default RHCOS.

https://redhat-internal.slack.com/archives/CUPJTHQ5P/p1705236817661129

docs/enhancements/0002_day0-integration.md
ybettan: By "us" I meant "KMM".

docs/enhancements/0002_day0-integration.md
ybettan: Never tries that. I would expect MCO to set an error in the `MachinecConfig`s statuses and not process any of them.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,282 @@
...
+Day-0 kmod are complex as they need to be present in the system at boot time,
+even before OCP is fully installed and certainly before KMM is deployed to the
+cluster, therefore, there is not much that KMM can do to help deploy those kmods but
+it can manage their lifecycle once a customer has installed a cluster with his
+day-0 kmods in it - this is what this enhancement is about.
+
+### Day-0 cluster background
+
+A customer wishing to deploy OCP with custom nodes will need to generate 2
+artifacts using [image-builder](https://www.redhat.com/en/topics/linux/what-is-an-image-builder)
```
romfreiman, line 20: does image builder capable already to generate a custom iso?

docs/enhancements/0002_day0-integration.md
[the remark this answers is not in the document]
ybettan: We won't. We will only need 1 `ModuleDay0`.

I understand your point. I will modify the `ModuleDay0` to target a MC and not a MCP.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+spec:
+  osMapping:
+    literal: "Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"
+    containerImage: quay.io/ybettan/rhcos:414.92.202312311229-0
+    build:
+      secrets:
+        - name: build-secret
+      dockerfileConfigMap:
+        name: kmm-kmod-dockerfile
+  machineConfigPoolSelector:
```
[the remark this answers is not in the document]
ybettan: We don't.
https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/files#r1451739088

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,285 @@
...
+The `ConfigMap` should be in the following format
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kmm-kmod-dockerfile
+data:
+  dockerfile: |
+
+    ARG DTK_AUTO
```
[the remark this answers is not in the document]
ybettan: In the same way it is obtained today.
It is explained later on: https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/files#diff-502e7d789feb94b276610d47444683bfd32820e1594f9413f0b310843f5682b4R190-R194

docs/enhancements/0002_day0-integration.md
romfreiman: backed -> baked

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,282 @@
...
+### Setting up the integration
+
+A user that wishes to hand over the management of his day-0 kmod to KMM will
+first need to make sure the `MachineConfig` can be reconize by KMM, therefore,
+he would add a `machineconfiguration.openshift.io/kmm-managed: "true"` label to
+it.
+
+KMM will watch for `MachineConfig`s in the cluster containing the
+`machineconfiguration.openshift.io/kmm-managed: "true"` label and create a new
+`ModuleDay0` CR in the cluster for each one of them.
```
romfreiman, line 63: why new cr is required rather that reuse an existing one? How conceptually it will be different?

docs/enhancements/0002_day0-integration.md
romfreiman: it does not make sense to maintain iso per node. it's not scalable.

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,282 @@
...
+the `MachineConfig` and the `node-role.kubernetes.io/worker: ""` label on the
+node.
+
+A customer can always create his own `MachineConfigPool` and add only 1 worker
+to it but I don't think this should be one of our primary concerns ATM for the
+following reasons
+  * assisted-installer will generate the same ISO for all nodes in the cluster
+    anyway so there is no benefit of having finer grained node selector for
+    `ModuleDay0` when assisted doesn't support such resolution on its own.
+  * if a user wishes to have only some worker nodes modified, he will need to
```
romfreiman: can u count how many reboots that cluster will go through untill it stabilizes to the target deployment?

docs/enhancements/0002_day0-integration.md
ybettan: No.

docs/enhancements/0002_day0-integration.md
ybettan: Conceptually they are very different.

In `Module` we do everything after a cluster upgrade (we build, sign, load etc) based on the new kernels we find.
In `ModuleDay0` we do everything before the cluster upgrade (build sign etc) because we need everything to be ready before the reboot.

Also, working with day0 mostly result in just updating the `MachineConfig` which is a very different thing than what we do for day2 operations such as:
1. Running `modprobe`
2. mounting some files to the node
3. etc

docs/enhancements/0002_day0-integration.md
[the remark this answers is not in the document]
ybettan: https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/commits/e4c6f6f7c8133336a025eba70fc630481d6f8fa3

docs/enhancements/0002_day0-integration.md
ybettan: numReboots=numWorkers-numNodesInTheNewPool

Anyway, I have removed this section from the proposal at it is MCO's internal implementation and not really related to the proposal

ybettan:
> General remark: I still think that KMM should not be the one responsible for updating the MachineConfig

@yevgeny-shnaidman If we put it in KMM then we have 2 options:
1. We add a flag that only register the new controller if explicitly requested 
    * Pros: we only need a single operator to manage both day2 and day0 kmods
    * Cons: operator becomes more complex
2. We create a new `kmm-day0` image (same as `kmm-hub`) for that controller
    * Pros: We keep the separation of concerns
    * Cons: If a user wishes to manage both day0 and day2 kmods he will need to install 2 operators

Do you still think it shouldn't be in KMM even if it is in a separate image?
If you think it doesn't belong here then let's schedule a meeting with @qbarrand, @mresvanis and @bthurber and decide where to put it.

docs/enhancements/0002_day0-integration.md
ybettan: Fixed. https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/commits/d0612e764912692cfe3c0653212252f407885b6e

docs/enhancements/0002_day0-integration.md
ybettan: > backed -> baked

Fixed. Thanks

docs/enhancements/0002_day0-integration.md
ybettan: The whole section was removed.

docs/enhancements/0002_day0-integration.md
ybettan: > it does not make sense to maintain iso per node. it's not scalable.

I agree. I just mentioned that to answer a potential question that says "but what if a customer really wan't to do it" - I have removed this whole section from the proposal.

docs/enhancements/0002_day0-integration.md
romfreiman: ok, so how will this enhancement help? :)

docs/enhancements/0002_day0-integration.md
romfreiman: and for prebuilt pre signed drivers, what CR would it be?

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,257 @@
...
+
+A user that wishes to hand over the management of his day-0 kmod to KMM will
+first need to make sure the `MachineConfig` can be reconize by KMM, therefore,
+he would add a `machineconfiguration.openshift.io/kmm-managed: "true"` label to
+it.
+
+KMM will watch for `MachineConfig`s in the cluster containing the
+`machineconfiguration.openshift.io/kmm-managed: "true"` label and create a new
+`ModuleDay0` CR in the cluster for each one of them.
+We will have a different controller for this new CRD in KMM.
```
romfreiman, line 64: I wonder if we can inject all those manifests during the installation, similar to what happens with the /var partition creation (lets understand how it works - because this is something that happens day0 vs day2):
https://docs.openshift.com/container-platform/4.12/installing/installing_bare_metal/installing-bare-metal.html#installation-user-infra-machines-advanced_vardisk_installing-bare-metal

docs/enhancements/0002_day0-integration.md
```diff
@@ -0,0 +1,257 @@
...
+`MachineConfig` with the new image, therefore, if a build/sign failed we won't
+reach the stage of modifying the `MachineConfig` at all and the cluster will
+remain valid.
+
+### Cluster upgrade
+
+When a `MachineConfig` uses `osImageURL`, it is called
+[image-layering](https://docs.openshift.com/container-platform/4.14/post_installation_configuration/coreos-layering.html).
+
+When image-layering is used, the OCP layer is being "detached" from the underlying
```
romfreiman, line 222: is it still the case? I was under the assumption that it should be addressed in future releases

docs/enhancements/0002_day0-integration.md
ybettan: Fixed as part of https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/commits/f666db49bc9b8ecac344c3b2d4a967ae60ad8e85

docs/enhancements/0002_day0-integration.md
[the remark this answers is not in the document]
ybettan: I have added an example https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/files#diff-502e7d789feb94b276610d47444683bfd32820e1594f9413f0b310843f5682b4R172-R179.
Hopefully, it helps.

docs/enhancements/0002_day0-integration.md
ybettan: The ISO building capability will reach image-builder/osbuild at some point :)

docs/enhancements/0002_day0-integration.md
ybettan: What "all manifests"? there is only the `MachineConfig` which is infact injected to the assisted-cluster pre-installation and the use can set the `machineconfiguration.openshift.io/kmm-managed: "true"` label pre-installation as well if he wishes. It won't have any effect until KMM is installed in the cluster though.

docs/enhancements/0002_day0-integration.md
ybettan: Me too. I haven't checked its status recently. Will ask.

ybettan:
@pcolledg-amd Thank you for your review.

> I'd suggest the RHCOS version string (ie. 414.92.202312311229-0) in ModuleDay0 CR be an internally-resolved templated variable a la KMM+DTK's ${KERNEL_FULL_VERSION}. Or even calculated from ${KERNEL_FULL_VERSION}.

Are you referring to the container image (`quay.io/ybettan/rhcos:414.92.202312311229-0`) or to the os literal (`"Red Hat Enterprise Linux CoreOS 414.92.202312311229-0 (Plow)"`)?

In KMM, we use those templates in case a user is targeting multiple nodes with different kernel using a regexp, therefore we need to template the kernel image to adjust it to each node. In the new day0 flow, a machineConfig will target nodes with the exact same OS on them so I am not sure I get the point of templating it.

docs/enhancements/0002_day0-integration.md
ybettan: It is still the same behavior indeed.
Looks like the [proposed-expriance](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/957/files#diff-502e7d789feb94b276610d47444683bfd32820e1594f9413f0b310843f5682b4R238-R248) may duplicate some effort planned in https://issues.redhat.com/browse/MCO-665.

@romfreiman Do you think we should avoid upgrading the nodes after a cluster upgrade due to the recent info about planned action in the MCO space?

romfreiman: i rather not have 2 mechanisms to achieve the same. And, it will introduce more reboots - 10m pare baremetal server.

docs/enhancements/0002_day0-integration.md
romfreiman: the question is whether that MC will become part of the ignition or only day2 (which affects number of reboots)

ybettan: I don't see any reference to the `MachineConfig` manifest in the discovery ignition. I am not sure when assisted applies the manifests that were added to the cluster pre-installation.

If that's answer your questions, that MC exists in the cluster as long as the cluster is customized and each change to it may trigger a reboot.

docs/enhancements/0002_day0-integration.md
ybettan: Then I will drop this "improvement" from the proposal.
FYI @qbarrand

ybettan:
After some discussions, we won't introduce an MCO-KMM integration for now, therefore, closing in favor of https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1004

/close

PR #958, opened 2024-01-11
Adding kernel module loading/unloading from tarball image file

yevgeny-shnaidman:
1. create cobra command that will support the kmod-tarball load/unload command line
2. add a MountImage implementation for tarball files
3. unit-test

### DIFF
--- a/cmd/worker/funcs_kmod.go
+++ b/cmd/worker/funcs_kmod.go
@@ -14,7 +14,6 @@ func rootFuncPreRunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to get appropriate ImageMounter: %v", err)
 	}
-
 	mr := worker.NewModprobeRunner(logger)
 	w = worker.NewWorker(im, mr, logger)
 
@@ -71,13 +70,30 @@ func setCommandsFlags() {
 		"",
 		"if set, this the value that firmware host path is mounted to")
 
+	kmodLoadCmd.Flags().Bool(
+		"tarball",
+		false,
+		"If true, extract the image from a tarball image instead of pulling from the registry",
+	)
+
 	kmodUnloadCmd.Flags().String(
 		worker.FlagFirmwareMountPath,
 		"",
 		"if set, this the value that firmware host path is mounted to")
+
+	kmodUnloadCmd.Flags().Bool(
+		"tarball",
+		false,
+		"If true, extract the image from a tarball image instead of pulling from the registry",
+	)
 }
 
 func getImageMounter(cmd *cobra.Command) (worker.ImageMounter, error) {
+	flag := cmd.Flags().Lookup("tarball")
+	if flag.Changed {
+		return worker.NewTarImageMounter(worker.ImagesDir, logger), nil
+	}
+
 	logger.Info("Reading pull secrets", "base dir", worker.PullSecretsDir)
 	keyChain, err := worker.ReadKubernetesSecrets(cmd.Context(), worker.PullSecretsDir, logger)
 	if err != nil {
--- a/internal/worker/tarimagemounter.go
+++ b/internal/worker/tarimagemounter.go
@@ -1,0 +1,52 @@
+package worker
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/go-logr/logr"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
+)
+
+type tarImageMounter struct {
+	ociImageHelper ociImageMounterHelperAPI
+	logger         logr.Logger
+	baseDir        string
+}
+
+func NewTarImageMounter(baseDir string, logger logr.Logger) ImageMounter {
+	ociImageHelper := newOCIImageMounterHelper(logger)
+	return &tarImageMounter{
+		ociImageHelper: ociImageHelper,
+		logger:         logger,
+		baseDir:        baseDir,
+	}
+}
+
+func (tim *tarImageMounter) MountImage(ctx context.Context, imagePath string, cfg *kmmv1beta1.ModuleConfig) (string, error) {
+	dstDir := filepath.Join(tim.baseDir, imagePath)
+
+	dstDirFS := filepath.Join(dstDir, "fs")
+	if err := os.MkdirAll(dstDirFS, os.ModeDir|0755); err != nil {
+		return "", fmt.Errorf("could not create the filesystem directory %s: %v", dstDirFS, err)
+	}
+
+	tim.logger.V(1).Info("Converting tarball into OCI image")
+
+	img, err := tarball.ImageFromPath(imagePath, nil)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


cmd/worker/funcs_kmod_tarball.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/worker"
+	"github.com/spf13/cobra"
+)
+
+var kmodTarballCmd = &cobra.Command{
+	Use:               "kmod-tarball",
+	Short:             "Manage kernel modules using tarball",
+	PersistentPreRunE: kmodTarballFuncPreRunE,
+}
```
qbarrand, line 14: Instead of this new command group, how about `worker kmod load-tarball` or `worker kmod load --tarball`?
I think `kmod` is still a good umbrella group for the new loading-from-tarball command.

cmd/worker/funcs_kmod_tarball.go
```diff
@@ -0,0 +1,61 @@
...
+
+	kmodTarballLoadCmd.Flags().String(
+		worker.FlagFirmwareMountPath,
+		"",
+		"if set, this the value that firmware host path is mounted to")
+
+	kmodTarballUnloadCmd.Flags().String(
+		worker.FlagFirmwareMountPath,
+		"",
+		"if set, this the value that firmware host path is mounted to")
```
qbarrand, line 60: I think we should try to reuse the function calls in `main` that apply to `kmodLoadCmd` and `kmodUnloadCmd`.
If we decide to go with `--tarball` instead, we can eliminate this new code entirely.

internal/worker/tarimagemounter.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (tim *tarImageMounter) MountImage(ctx context.Context, imagePath string, cfg *kmmv1beta1.ModuleConfig) (string, error) {
+	dstDir := filepath.Join(tim.baseDir, imagePath)
+
+	dstDirFS := filepath.Join(dstDir, "fs")
+	if err := os.MkdirAll(dstDirFS, os.ModeDir|0755); err != nil {
+		return "", fmt.Errorf("could not create the filesystem directory %s: %v", dstDirFS, err)
+	}
+
+	tim.logger.V(1).Info("Pulling image")
```
qbarrand, line 37: We're not really pulling here.

internal/worker/tarimagemounter.go
```diff
@@ -0,0 +1,52 @@
...
+
+	img, err := tarball.ImageFromPath(imagePath, nil)
+	if err != nil {
+		return "", fmt.Errorf("could not create image from path  %s: %v", imagePath, err)
+	}
+
+	err = tim.ociImageHelper.mountOCIImage(img, dstDirFS)
+	if err != nil {
+		return "", fmt.Errorf("failed mounting oci image: %v", err)
```
qbarrand, line 46: I am really not a fan of "mounting", I think it might be confusing to users. We should fix that upstream and change it accordingly.

cmd/worker/funcs_kmod_tarball.go
yevgeny-shnaidman: there are a couple of arguments for using new command group:
1. we can initialize worker in the kmod and kmod-tarball functions, instead of initializing them with ifs in each of the Load/Unload commands
2. it gives us an alost complete separation between kmod and kmod-tarball code , which is great in terms of backporting.  Right now we have exactly one line that differs between upstream and downstream code.

cmd/worker/funcs_kmod_tarball.go
yevgeny-shnaidman: i can add a re-use function in case we agree on the remark above

internal/worker/tarimagemounter.go
yevgeny-shnaidman: fixed

internal/worker/tarimagemounter.go
yevgeny-shnaidman: IMHO "mount" in this context is more clear than "extract", but i don't have a strong opinions regarding naming. In any case i agree that it should be first done in the upstream. We can add it to the "re-factoring" task for v2

cmd/worker/funcs_kmod.go
```diff
@@ -64,3 +59,56 @@ func kmodUnloadFunc(cmd *cobra.Command, args []string) error {
...
+	klogFlagSet := flag.NewFlagSet("klog", flag.ContinueOnError)
+	logConfig := textlogger.NewConfig()
+	logConfig.AddFlags(klogFlagSet)
+	logger = textlogger.NewLogger(logConfig).WithName("kmm-worker")
+	rootCmd.PersistentFlags().AddGoFlagSet(klogFlagSet)
+}
+
+func getImageMounter(cmd *cobra.Command) (worker.ImageMounter, error) {
+	flag := cmd.Flags().Lookup("tarball")
+	if flag.Changed {
```
qbarrand, line 103: Should we instead just get the value here?

cmd/worker/funcs_kmod.go
```diff
@@ -64,3 +59,56 @@ func kmodUnloadFunc(cmd *cobra.Command, args []string) error {
...
+	)
+}
+
+func configureLogging() {
+	klogFlagSet := flag.NewFlagSet("klog", flag.ContinueOnError)
+	logConfig := textlogger.NewConfig()
+	logConfig.AddFlags(klogFlagSet)
+	logger = textlogger.NewLogger(logConfig).WithName("kmm-worker")
+	rootCmd.PersistentFlags().AddGoFlagSet(klogFlagSet)
+}
```
qbarrand, line 99: Why do we need this as a separate function?

cmd/worker/funcs_kmod.go
```diff
@@ -64,3 +59,56 @@ func kmodUnloadFunc(cmd *cobra.Command, args []string) error {
...
+
+func getImageMounter(cmd *cobra.Command) (worker.ImageMounter, error) {
```
qbarrand, line 101: Why do we need this as a separate function?

cmd/worker/funcs_kmod.go
yevgeny-shnaidman: it can be both ways. It is a boolean , with default as "false", so "Changed" indicates true

cmd/worker/funcs_kmod.go
yevgeny-shnaidman: just to make a code more cleaner

cmd/worker/funcs_kmod.go
```diff
@@ -64,3 +59,56 @@ func kmodUnloadFunc(cmd *cobra.Command, args []string) error {
...
+
+	kmodLoadCmd.Flags().String(
+		worker.FlagFirmwareMountPath,
+		"",
+		"if set, this the value that firmware host path is mounted to")
+
+	kmodLoadCmd.Flags().Bool(
+		"tarball",
+		false,
+		"if see, means run tarball",
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

qbarrand:
/lgtm
/hold
@yevgeny-shnaidman please feel free to unhold when you're happy.

PR #965, opened 2024-01-15
re-factoring worker's main and funcs_kmods (#691)

yevgeny-shnaidman:


yevgeny-shnaidman:
Fixes [964](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/964)

PR #968, opened 2024-01-16
re-fixing DevicePlugin upgrade from v1.1 to v2

yevgeny-shnaidman:
Unlike upstream, in downstream version ModuleLoader daemonset differs from DevicePlugin by present of the kernel version label, and not by role label.
This PR changes the identification of DevicePlugins during gathering all device plugin DS by checkig the absence of the kernel version label

### DIFF
--- a/internal/controllers/device_plugin_reconciler.go
+++ b/internal/controllers/device_plugin_reconciler.go
@@ -176,7 +176,7 @@ func (dprh *devicePluginReconcilerHelper) getModuleDevicePluginDaemonSets(ctx co
 	devicePluginsList := make([]appsv1.DaemonSet, 0, len(dsList.Items))
 	// remove the older version module loader daemonsets
 	for _, ds := range dsList.Items {
-		if ds.GetLabels()[constants.DaemonSetRole] != constants.ModuleLoaderRoleLabelValue {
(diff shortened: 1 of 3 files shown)


internal/controllers/device_plugin_reconciler.go
```diff
@@ -176,7 +176,7 @@ func (dprh *devicePluginReconcilerHelper) getModuleDevicePluginDaemonSets(ctx co
 	devicePluginsList := make([]appsv1.DaemonSet, 0, len(dsList.Items))
 	// remove the older version module loader daemonsets
 	for _, ds := range dsList.Items {
-		if ds.GetLabels()[constants.DaemonSetRole] != constants.ModuleLoaderRoleLabelValue {
+		if _, ok := ds.GetLabels()[constants.KernelLabel]; !ok {
```
ybettan, line 179: Why do we have that diff in the labels of the DS between u/s and d/s?

qbarrand: Because we released upstream before we tested downstream. That is why now we wait for testing to be completed downstream before releasing upstream.
BTW this means that the 1.0 --> 1.1 upgrade was probably broken upstream.

PR #1004, opened 2024-02-08
Add documentation on how day0/day1 kmods can be managed with KMM.

ybettan:
Some kmods might need to be installed as part of the cluster installation.

In some cases, we can handover the lifecycle management of those kmods to KMM - this is what this doc is about.

### DIFF
--- a/docs/mkdocs/documentation/day0_day1_handover.md
+++ b/docs/mkdocs/documentation/day0_day1_handover.md
@@ -1,0 +1,86 @@
+# Managing Day0/Day1 kmods with KMM
+
+Some kmods might be installed without KMM. In order to enhance KMM's UX we
+could, in some cases, help customers to transition the lifecycle management
+of they kmods to KMM.
+
+## Definitions
+
+### Day 0
+
+The most basic kmods that are required for a node to become “Ready” in the cluster
+
+**Examples**
+* A storage driver that is required in order to mount the rootFS as part of the boot process.
+  Vendors will usually work closely with the RHEL team to make those drivers
+  in-tree so we won’t worry about them too much here.
+* A network driver that is required for the machine to access machine-config-server
+  on the bootstrap node to pull the ignition and join the cluster
+
+### Day 1
+
+Kmods that are not required for a node to become “Ready” in the cluster but would
+not be able to be unloaded once the node is "Ready".
+
+**Examples**
+* An OOT network driver that replaces an outdated in-tree driver to exploit the
+  full potential of the NIC while `NetworkManager` depends on it.
+  Once the node is "Ready" a customer won't be able to unload the driver because
+  of the `NetworkManager` dependency.
+
+### Day2
+
+Kmods that can be dynamically loaded to the kernel or removed from it without
+interfering with the cluster infrastructure (such as connectivity).
+
+**Examples**
+* GPU operator
+* Secondary network adapters
+* FPGA
+
+## Layering background
+
+When a day0 kmod was installed in the cluster, it means that “layering” was applied
+through MCO and OCP upgrades won’t trigger node upgrades.
+
+Unless a user wants to add new features to its driver, we will never need to
+recompile it for them since the node’s OS will remain.
+
+With that being said, MCO has [plans](https://issues.redhat.com/browse/MCO-665)
+to rebuild the node images upon a cluster upgrade when Layering is used by MCO.
+
+## Using KMM for managing day0 and day1 kmods
+
+We can leverage KMM to manage the lifecycle of day0/1 kmods without a reboot when the driver allows it.
+NOTE: It will not work if the upgrade require a node reboot (when rebuilding initramfs is needed for example)
+
+### 1st option
+
+By treating the kmod as an in-tree driver.
+
+Nothing to do until user wishes to update the kmods.
+
+When the user wishes to upgrade the kmod, they treat it as an in-tree driver
+and create a `Module` in the cluster with the `inTreeRemoval` field to unload
+the old version of the driver.
+
+**Characteristics**
+
+* Down time - KMM will try to unload and load the kmod on all the selected nodes simultaneously.
+* Works in case removing the driver makes the node lose connectivity (because KMM uses a single pod to unload+load the driver)
+
+### 2nd option
+
+By using ordered upgrade.
+
+In this case, user creates a versioned `Module` in the cluster representing the kmods - nothing
+will happen since the kmods are already loaded.
+
+When the user wishes to upgrade the kmod, they use the ordered-upgrade feature.
+
+**Characteristics**
+
+* No cluster downtime - the user controls the pace of the upgrade, and how many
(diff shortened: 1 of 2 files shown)


docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+* A storage driver that is required in order to mount the rootFS as part of the boot process.
+  Vendors will usually work closely with the RHEL team to make those drivers
+  in-tree so we won’t worry about them too much here.
+* A network driver that is required for the machine to access machine-config-server
+  on the bootstrap node to pull the ignition and join the cluster
+
+### Day 1
+
+Kmods that are not required for a node to become “Ready” in the cluster but are
+preferred to be installed/updated before the node is becoming “Ready”.
```
qbarrand, line 23: If the kmod is not required to be loaded before the node is `Ready`, then I think we should rather mention that the kmod is a requirement for a certain application.

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Examples**
+* An OOT network driver that replaces an outdated in-tree driver to exploit the full potential of the NIC.
+
+### Day2
+
+Kmods that can be dynamically loaded to the kernel or removed from it without
+interfering with the cluster infrastructure (such as connectivity).
```
qbarrand, line 31: Already the case with day 1, no?

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+## Layering background
+
+When a day0 kmod was installed in the cluster, it means that “layering” was applied
+through MCO and OCP upgrades won’t trigger node upgrades.
+
+Unless a user wants to add new features to its driver, we will never need to
+recompile it for him since the node’s OS will remain.
+
+With that being said, MCO has [plans](https://issues.redhat.com/browse/MCO-665)
+to rebuild the node images upon a cluster upgrade when Laying is used by MCO.
```
qbarrand, line 47:
```suggestion
to rebuild the node images upon a cluster upgrade when Layering is used by MCO.
```

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+
+We can leverage KMM to manage the lifecycle of day0/1 kmods without a reboot when the driver allows it.
+
+### 1st option
+
+By treating the kmod as an in-tree driver.
+
+Nothing to do until user wishes to update the kmods.
+
+When user wishes to upgrade the kmod, he treats the kmod as an in-tree driver
```
qbarrand, line 59:
```suggestion
When the user wishes to upgrade the kmod, they treat it as an in-tree driver
```

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+and creates a `Module` in the cluster with the `intreeRemoval` field to unload
```
qbarrand, line 60:
```suggestion
and create a `Module` in the cluster with the `inTreeRemoval` field to unload
```

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+* Doesn’t work if the upgrade require a node reboot (when rebuilding initramfs is needed for example)
+
+### 2nd option
+
+By using ordered upgrade.
+
+In this case, user creates a versioned `Module` in the cluster representing the kmods - nothing
+will happen since the kmods are already loaded.
+
+When user wishes to upgrade the kmod, he uses the ordered-upgrade feature.
```
qbarrand, line 76:
```suggestion
When the user wishes to upgrade the kmod, they use the ordered-upgrade feature.
```

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Characteristics**
+
+* No cluster downtime - KMM will make the upgrade node by node
```
qbarrand, line 80: Rather - the user controls the pace of the upgrade, and how many nodes are upgraded at the same time.

docs/mkdocs/documentation/day0_day1_handover.md
```diff
@@ -0,0 +1,83 @@
...
+* Doesn't work if unloading the driver results in losing connection to the node
+  (because KMM will create 2 different worker pods, one for unloading and another for loading which won’t be scheduled)
+* Doesn’t work if the upgrade require a node reboot (when rebuilding initramfs is needed for example)
```
qbarrand, line 83: I think it does.
- KMM unloads old driver
- User reboots node
- Node boots with the old-old driver
- KMM loads new driver

docs/mkdocs/documentation/day0_day1_handover.md
ybettan: It is not always the case. For example, an outdated in-tree driver that is replaces by an up-to-date OOT network driver:
The node can become ready without the driver, but performance can be better with the new driver.
So it is not required for the infra, nor an up, but it improves the infra connection.

docs/mkdocs/documentation/day0_day1_handover.md
ybettan: Not exactly. Let's take the day1 use case that @yevgeny-shnaidman has worked on - an up-do-date OOT network driver that replaces an outdated in-tree network driver - it cannot be dynamically unloaded from the node without breaking the node's connection.

docs/mkdocs/documentation/day0_day1_handover.md
ybettan: Fixed.

docs/mkdocs/documentation/day0_day1_handover.md
ybettan: You have a few assumption here (which doesn't always line up)
1. If it's a day0 kmod, you don't have an "old-old driver" - without the new driver - the node won't connect to the cluster
2. You assume that the kmod can be loaded with `modprobe` as a day2 operation - this is not the case with day0 and some day1 kmods.

If you need the driver in the initramFS image, there is no other option, than to re-build the initramFS image, let MCO know about the change and reboot with the new initramFS image (KMM doesn't know how to handle any of those steps)

ybettan: If it's a day0 kmod you don't have any "old-old driver" in the node's image.

qbarrand: Do you need to have it?

docs/mkdocs/documentation/day0_day1_handover.md
ybettan: Fixed.

PR #1022, opened 2024-02-15
Adding an operator-upgrade CI job. (#693)

ybettan:
This test will ensure the bundle built in the PR can be used to upgrade from the last bundle available of KMM.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/991
/cc @qbarrand

### DIFF
--- a/ci/prow/operator-upgrade
+++ b/ci/prow/operator-upgrade
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+make operator-sdk
+oc create namespace openshift-kmm
+
+# Set the pull secret in Docker's auth file so operator-sdk can pull the current bundle image
+mkdir -p ~/.docker
+oc get secret/pull-secret -n openshift-config -o json | jq -r '.data[".dockerconfigjson"]' | base64 --decode | jq > ~/.docker/config.json
+
+# Get the latest bundle image published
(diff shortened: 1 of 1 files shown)


ci/prow/operator-upgrade
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# Get the latest bundle image published
+latest_published_bundle=$(curl -sSL "https://datagrepper.engineering.redhat.com/raw?topic=/topic/VirtualTopic.eng.ci.redhat-container-image.index.built&delta=2592000&contains=kernel-module-management-operator-bundle-container" | jq '.raw_messages[0].msg.index.index_image')
```
qbarrand, line 6: If we are releasing for two streams, say `1.y` and `2.y`, would it be possible that this returns a `1.y` bundle?

ybettan:
/hold
To be merged after https://github.com/openshift/release/pull/48838 for validation.

ybettan:
https://github.com/openshift/release/pull/48838 was merged.

/retest

ybettan:
> @ybettan any idea why we keep seeing this?

Yes, I am still working on it.

Since `clone: true` isn't supported in multi-steps jobs but cluster claims requires multi-steps jobs, I had to use another image based on `src` in CI.

The current issue is that our job needs access to the pull secret of the `registry.redhat.io` registry. See this [slack thread](https://redhat-internal.slack.com/archives/CBN38N3MW/p1709188619467149).

/hold

PR #1040, opened 2024-03-06
Fixing day1 scenario to work with v2

yevgeny-shnaidman:
1) worker image shoudl also be pulled in the
   pull-kernel-module-image.service in order to be present on the node
   when the replace-kernel-module.service is running and network has not
   been configured yet
2) adding option to the API to provide the worker image and to provide
   the in-tree kernel module to be removed. Previously the loaded kernel
   module and the in-tree kernel module had to be the same ones.
3) changing the day-utility to support the new API
4) updating unit-test


cmd/day1-utility/main.go
```diff
@@ -14,27 +14,36 @@ var (
...
+		kernelModule       string
+		workerImage        string
+		inTreeKernelModule string
+	)
+	flag.StringVar(&mcName, "machine-config", "", "name of the machine config to create")
+	flag.StringVar(&mpName, "machine-config-pool", "", "name of the machine config pool to use")
+	flag.StringVar(&kernelModuleImage, "image", "", "container image that contains kernel module .ko file")
+	flag.StringVar(&kernelModule, "kernel-module", "", "container image that contains kernel module .ko file")
+	flag.StringVar(&workerImage, "worker-image", "", "kernel-management worker image to use. If not passed, a default value will be used")
+	flag.StringVar(&inTreeKernelModule, "in-tree-module-to-remove", "", "in-tree kernel module that should be removed prior to loading the oot module")
```
qbarrand, line 41: Why are the flags back? Didn't we agree that mandatory values should be arguments, not flags?

yevgeny-shnaidman: Since i have started using this utility more extensively, i found out that using flags is actually more convenient and less error prone

PR #1041, opened 2024-03-06
changing inTreeRemoval to support multiple modules (#747)

yevgeny-shnaidman:
This PR contains the following changes
1) Adding new field InTreeModulesToRemove to the Module that is of type string array 2) Marking the old field InTreeModuleToRemove as deprecated 3) adding validation to webhook that InTreeModuleToRemove and InTreeModulesToRemove
   can not be both set in the Module
4) appropriate changes are done in the MLD structures and in the worker
   configuration
5) for older version Modules that already use InTreeModuleToRemove
   field, its value is converted into string array in ModuleLoaderData
6) adding InTreeModulesToRemove to the ModuleConfig of the NMC 7) in case NMC was already set with InTreeModuleToRemove prior to update
   of the KMM, worker will use InTreeModuleToRemoce in case InTreeModuleToRemove
   is not set
8) code and unit-tests are updated


qbarrand:
MGMT-16638

Issue #898, opened 2023-11-17
KMM Day1 should support multiple modules


yevgeny-shnaidman:
Currently KMM Day functionality supports replacing only one kernel module during linux kernel initialization.
We should allow support for replacing multiple modules during kernel initialization.
There are 2 possible implementation option:
1) All the module are stored in the same image. The systemd service will be changed to go over a list of modules and replace each on of them.  This will require changing the package API implementation to receive a list of modules
2) Each such kernel module should be replaced using a different MachineConfig. This will require making sure that the systemd for each such MC are not named and located at the same place, in order not to override one another

The first option  seems more feasible, since it won't require user to create multiple images, and will require only one call to the package API

PR #1055, opened 2024-03-18
day1 cleanup after the action

yevgeny-shnaidman:
pull-kernel-module-image.service runs twice: once to pull all the images and reboot the node, and the second time it just checks that everyting has been pulled and exits. During the second time we want to remove the tar of the loaded and saved kernel module image, so that it won't interfere with the correct flow once the node is rebooted again


qbarrand:
Don't we want to keep the tarfile so that we don't have to pull it again on next boot?

yevgeny-shnaidman:
> Don't we want to keep the tarfile so that we don't have to pull it again on next boot?

I prefer to use imagePull "Always" policy. Currently the tar file name is always the same, so just to be sure we delete it and pull everything a new. When i get to the [issue 898]( https://github.com/rh-ecosystem-edge/kernel-module-management/issues/898 ) and [issue 920](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/920) i will probably change tar file name.
But it is still better to delete it, since it also indicates the presence of the worker image. If for some reason worker image got GCed between reboots, we want to pull it again

PR #1066, opened 2024-03-28
Adding support for on-prem OCP clusters to day1 scenario

yevgeny-shnaidman:
Adding support for on-prem OCP clusters to day1 scenario
    
    On-prem OCP cluster, installed via assisted-installed, has an
    additional service that is resposible for configuring DNS on host.
    Therefore usual day1 service, that is dependent only on NM service, will
    be unable to pull images prior to DNS configuration service being
    completed. Since systems service do not support "if else" syntax
    we will define 2 pull-kernel-module-image services:
    1) service that is supposed to be run on usual cluster will be executed
       only in case /etc/systemd/system/on-prem-resolv-prepender.service
       file is missing from the host
    2) service that is supposed to be run on on-prem clusters will run only
       in case /etc/systemd/system/on-prem-resolv-prepender.service is
       present on the host, and will depend both on NM service and on
       NetworkManager-dispatcher service
    
    Since NetworkManager-dispatcher service is a dbus service, the second
    service should verify if it has completed successfully before running
    the main executable. This is acomplished by a dedicated script that wait
    on NetworkManager-dispatcher service and runs as a PresStart script

### DIFF
--- a/pkg/mcproducer/mcproducer.go
+++ b/pkg/mcproducer/mcproducer.go
@@ -30,6 +30,10 @@ var (
 		template.ParseFS(templateFS, "templates/replace-kernel-module.gotmpl"),
 	)
 
+	scriptWaitForNetworkDispatcher = template.Must(
+		template.ParseFS(templateFS, "templates/wait-for-dispatcher.gotmpl"),
+	)
+
 	workerConfigMap = template.Must(
 		template.ParseFS(templateFS, "templates/worker-configmap.gotmpl"),
 	)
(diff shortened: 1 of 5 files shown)


pkg/mcproducer/testdata/machineconfig-test.yaml
```diff
@@ -32,8 +32,9 @@ spec:
         - contents: |
             [Unit]
             Description=Pull oot kernel module image
```
qbarrand, line 34: Should have different description for both services?

pkg/mcproducer/mcproducer.go
```diff
@@ -71,6 +75,11 @@ func ProduceMachineConfig(machineConfigName,
 		return "", err
 	}
 
+	templateParams["WaitForNetworkDispatcherContents"], err = executeIntoBase64(scriptWaitForNetworkDispatcher, templateParams)
```
qbarrand, line 78: Does this need to be rendered as a template? I don't see any variable in it, we could probably just encode it to base64.

yevgeny-shnaidman: Yes, but we already have a function for encoding, so even if there are no template variables in it, i prefer to use existing function, then defining a new one, or calling an base64 API directly

pkg/mcproducer/testdata/machineconfig-test.yaml
yevgeny-shnaidman: fixed

PR #1069, opened 2024-04-01
making day1 solutin configurable via systemd services

yevgeny-shnaidman:
1) adding environment variables to systemd services and making executable
   using those variables. This ways the scripts themselves do not change.
   in addition, it allows manual update of the MC in the cluster, since
   no parts that are based64-ed are needed to be changed
2) removing a separate pulling service for OCP on-prem. pull service
   will now wait for network-online and NetworkManager-dispatcher to be
   completed in every type of OCP cluster


pkg/mcproducer/templates/pull-image.sh
qbarrand, line 1: Should we move this to another directory as it's not a template?

pkg/mcproducer/templates/replace-kernel-module.sh
qbarrand, line 1: Should we move this to another directory as it's not a template?

pkg/mcproducer/templates/replace-kernel-module.sh
```diff
@@ -0,0 +1,35 @@
...
+kernel_module_image_filepath="$KERNEL_MODULE_IMAGE_FILEPATH"
+in_tree_module_to_remove="$IN_TREE_MODULE_TO_REMOVE"
+kernel_module="$KERNEL_MODULE"
+worker_image="$WORKER_IMAGE"
+
+create_kmm_config() {
+    # Write YAML content to the file
+    cat <<EOF > "$kmm_config_file_filepath"
+containerImage: $kernel_module_image_filepath
+inTreeModuleToRemove: $in_tree_module_to_remove
```
qbarrand, line 13: I think this variable is not mandatory, so we should quote it to prevent the YAML from being unparsable if `$in_tree_module_to_remove` is an empty string.

pkg/mcproducer/testdata/machineconfig-test.yaml
```diff
@@ -20,6 +20,11 @@ spec:
             User=root
             Type=oneshot
             TimeoutSec=10
+            Environment="KERNEL_MODULE_IMAGE_FILEPATH=/var/lib/image_file_day1.tar"
+            Environment="WORKER_IMAGE=quay.io/edge-infrastructure/kernel-module-management-worker:latest"
+            Environment="WORKER_CONFIG_FILEPATH=/var/lib/kmm_day1_config.yaml"
+            Environment="IN_TREE_MODULE_TO_REMOVE=testInTreeKernelModuleName"
+            Environment="KERNEL_MODULE=testKernelModuleName"
```
qbarrand, line 27: Nit: shall we keep the alphabetical order here?

pkg/mcproducer/testdata/machineconfig-test.yaml
```diff
@@ -32,31 +37,15 @@ spec:
...
-            ConditionPathExists=/etc/systemd/system/on-prem-resolv-prepender.service
             After=network-online.target
             After=NetworkManager-dispatcher.service
             DefaultDependencies=no
             [Service]
             User=root
             Type=oneshot
+            Environment="KERNEL_MODULE_IMAGE_FILEPATH=/var/lib/image_file_day1.tar"
+            Environment="WORKER_IMAGE=quay.io/edge-infrastructure/kernel-module-management-worker:latest"
+            Environment="KERNEL_MODULE_IMAGE=quay.io/project/repo:some-tag12"
```
qbarrand, line 48: Nit: shall we keep the alphabetical order here?

pkg/mcproducer/testdata/pull-image-test.sh
```diff
@@ -1,31 +1,36 @@
...
-    echo "File /var/lib/image_file_day1.tar found.Nothing to do, the file was handled, removing it"
-    rm -f /var/lib/image_file_day1.tar
+
+kernel_module_image_filepath="$KERNEL_MODULE_IMAGE_FILEPATH"
+worker_image="$WORKER_IMAGE"
+kernel_module_image="$KERNEL_MODULE_IMAGE"
+
+if [ -e $kernel_module_image_filepath ]; then
+    echo "File $kernel_module_image_filepath found.Nothing to do, the file was handled, removing it"
+    rm -f $kernel_module_image_filepath=
```
qbarrand, line 10:
```suggestion
    rm -f $kernel_module_image_filepath
```

pkg/mcproducer/testdata/pull-image-test.sh
qbarrand, line 1: Do we need to keep this test file?

pkg/mcproducer/testdata/replace-kernel-module-test.sh
qbarrand, line 1: Do we need to keep this test file?

pkg/mcproducer/templates/pull-image.sh
yevgeny-shnaidman: the scripts are not template, but the machine-config is. We can create 2 directories: template and execs, but i am not sure if it is more clear. But i am not hard-locked on that, if you are adamant about it, i can change it

pkg/mcproducer/templates/replace-kernel-module.sh
yevgeny-shnaidman: actually, yaml does not become unparsable, i have verified it. This way i don't need to have additional ifs in bash script.

pkg/mcproducer/testdata/machineconfig-test.yaml
yevgeny-shnaidman: fixed

pkg/mcproducer/testdata/machineconfig-test.yaml
yevgeny-shnaidman: fixed

pkg/mcproducer/testdata/pull-image-test.sh
yevgeny-shnaidman: fixed

pkg/mcproducer/testdata/pull-image-test.sh
yevgeny-shnaidman: i think it makes it clearer on what data should the machine-config-test be based.

pkg/mcproducer/testdata/replace-kernel-module-test.sh
yevgeny-shnaidman: same as above

pkg/mcproducer/templates/pull-image.sh
qbarrand: I would indeed move the embedded files to another directory outside of `templates`.

pkg/mcproducer/testdata/pull-image-test.sh
qbarrand: But we're not rendering it as a template anymore, so does it make sense to check its contents?

pkg/mcproducer/templates/replace-kernel-module.sh
qbarrand: TIL.

pkg/mcproducer/templates/replace-kernel-module.sh
yevgeny-shnaidman: please see the answer above

pkg/mcproducer/templates/pull-image.sh
yevgeny-shnaidman: fixed

pkg/mcproducer/testdata/replace-kernel-module-test.sh
yevgeny-shnaidman: fixed

pkg/mcproducer/templates/replace-kernel-module.sh
yevgeny-shnaidman: fixed

PR #1072, opened 2024-04-03
e2e-hub: install in different namespaces

qbarrand:
In https://github.com/openshift/release/pull/49549, we want to transition to bundles for those CI jobs and operator-sdk will refuse to install both operators in the same OperatorGroup because they have different install modes.

/cc @ybettan


ci/e2e/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM ${DTK_AUTO} as builder
+
+ARG KERNEL_VERSION
+
+WORKDIR /usr/src
+
+RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret
+
+RUN git clone https://github.com/rh-ecosystem-edge/kernel-module-management.git
```
ybettan, line 11: Can we use `COPY` with the local file instead of `RUN git clone`?

PR #1119, opened 2024-05-21
MGMT-17820: Do not watch Build objects when managed

qbarrand:


yevgeny-shnaidman:
looks ok, i will lgtm once all the tests have passed

PR #1168, opened 2024-08-08
Set EnvironmentFile=/etc/mco/proxy.env in the systemd services

borball:
Set EnvironmentFile=/etc/mco/proxy.env in the systemd services so that the script will respect the proxy settings on the cluster.

### DIFF
--- a/pkg/mcproducer/templates/machine-config.gotmpl
+++ b/pkg/mcproducer/templates/machine-config.gotmpl
@@ -20,6 +20,7 @@ spec:
             User=root
             Type=oneshot
             TimeoutSec=10
+            EnvironmentFile=/etc/mco/proxy.env
             Environment="IN_TREE_MODULE_TO_REMOVE={{.InTreeModuleToRemove}}"
             Environment="KERNEL_MODULE={{.KernelModule}}"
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
why is it needed?

borball:
@yevgeny-shnaidman on a cluster which has proxy enabled we need that file to enable the proxy so that the scripts can fetch the images, otherwise the image pulling will be failed.

yevgeny-shnaidman:
> @yevgeny-shnaidman on a cluster which has proxy enabled we need that file to enable the proxy so that the scripts can fetch the images, otherwise the image pulling will be failed.

Ok, i see, thank you for explanation! I am ok with the PR, please just fix the unit-tests, and we can push it

ybettan:
/retest
@borball Thank you for your PR.
Please rebase and squash your PR into a single commit with a descriptive commit message that describe why it is needed.

We have a [CI job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/rh-ecosystem-edge_kernel-module-management/1168/pull-ci-rh-ecosystem-edge-kernel-module-management-main-check-commits-count/1826788073334116352) enforcing a single commit per PR to prevent merging PRs with commit messages like `fix unit test`:

borball:
@ybettan Thanks, I fixed the commit message.

PR #1183, opened 2024-08-20
Simplifying firmware flags between NMC controller and worker pod

yevgeny-shnaidman:
Currently NMC controller passes 2 flag to worker pod: firmware path mount point in pod and firmware search path. Since in our implementation those 2 values are always equal, we can use only one flag to pass these values: --firmware-path, and let worker pod decide when to update the kernel configuration for firmware search


ybettan:
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1173

ybettan:
/retest
`NoClusters: No clusters in pool are ready to be claimed`

PR #1186, opened 2024-08-28
Fixing some `kustomizeconfig.yaml` files.

ybettan:
Since `kustomize@v5.4.0`, `yaml.UnmarshalStric()` is used instead of `yaml.Unmarshal()` which means that fields in the yamls that doesn't exist in the golang object will fail to unmarshal instead of being ignored like they would without the `strict` behavior.

Therefore, those redundant `name` fields where removed.

In addition, I have bumped `kustomize@5.3.0` --> `kustomize@5.4.3` to make sure we get a closer behavior on machines that comes with `kustomize` pre-installed and those which aren't and install it on-the-fly.

I didn't see any diff in the output of `kustomize build config/default` with `kustomize@5.3.0` and the redundant `name` fields and with `kustomize@5.4.3` without them.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1184

/assign @yevgeny-shnaidman 
/cc @TomerNewman

### DIFF
--- a/config/default-hub/kustomizeconfig.yaml
+++ b/config/default-hub/kustomizeconfig.yaml
@@ -1,4 +1,3 @@
 namePrefix:
 - kind: Deployment
-  name: controller
   path: spec/template/spec/volumes/secret/secretName
--- a/config/default/kustomizeconfig.yaml
+++ b/config/default/kustomizeconfig.yaml
@@ -1,4 +1,3 @@
 namePrefix:
 - kind: Deployment
-  name: controller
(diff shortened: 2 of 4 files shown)


ybettan:
/retest

Looks like a cluster installation issue:
```
msg="Failed to run bundle: create catalog: error creating registry pod: registry pod did not become ready: error waiting for registry pod ff4ba46fa11f8b2ebcc27212f21042f5e18997e6091ed7283021d53ca8beb87 to run: timed out waiting for the condition\n"
```

PR #1188, opened 2024-08-28
Day1 feature support multiple kmod

TomerNewman:
Previously, the `replace-kernel-module.service` and `pull-kernel-module-image.service` files were created without differentiation for each kernel module, leading to potential overlap. This update introduces a prefix using <kmod_name>, resulting in service files named `<kmod_name>-replace-kernel-module.service` and `<kmod_name>-pull-kernel-module-image.service`, thereby resolving the issue.

---

/assign @yevgeny-shnaidman
/cc @ybettan

### DIFF
--- a/pkg/mcproducer/templates/machine-config.gotmpl
+++ b/pkg/mcproducer/templates/machine-config.gotmpl
@@ -34,7 +34,7 @@ spec:
             [Install]
             WantedBy=multi-user.target
           enabled: true
-          name: "replace-kernel-module.service"
+          name: "{{.MachineConfigName}}-replace-kernel-module.service"
         - contents: |
             [Unit]
             Description=Pull oot kernel module image
@@ -56,7 +56,7 @@ spec:
             [Install]
             WantedBy=multi-user.target
           enabled: true
-          name: "pull-kernel-module-image.service"
+          name: "{{.MachineConfigName}}-pull-kernel-module-image.service"
         - enabled: false
           mask: true
           name: crio-wipe.service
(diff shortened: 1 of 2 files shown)


pkg/mcproducer/templates/machine-config.gotmpl
```diff
@@ -34,7 +34,7 @@ spec:
             [Install]
             WantedBy=multi-user.target
           enabled: true
-          name: "replace-kernel-module.service"
+          name: "{{.KernelModule}}-replace-kernel-module.service"
```
yevgeny-shnaidman, line 37: I thought we would be using the MachineConfig name as the prefix... WDYT?

TomerNewman: > I thought we would be using the MachineConfig name as the prefix... WDYT?

@yevgeny-shnaidman  but the override can happen only when we try more than 1 kmod, not? like it says on #920

TomerNewman: fixed

yevgeny-shnaidman: @TomerNewman have you run it on a test cluster?

TomerNewman: > @TomerNewman have you run it on a test cluster?

I ran unit tests + ran it locally like we did on the meeting.
I will try then to ran it on a cluster additionally

yevgeny-shnaidman: the test is to create a simple container image with simple kmod, create the yaml, and then apply it to a cluster and see of the kernel module is loaded into the kernel

TomerNewman: done, the module was loaded into the kernel at my test cluster

PR #1210, opened 2024-09-11
support inTree removal scenario for nodes without .ko files

yevgeny-shnaidman:
Modprobe will fail with error in case it cannot find an appropriate .ko file when asked to remove a kernel module. With in-tree removal scenario, modporbe search path is /lib/modules on the host, and not the worker image.
This PR contains the following:
1) Adding FileExists function that check the presence of a file under
   a search path based on a regexp
2) in worker flow, in case in-tree removal scenario is requisted, worker
   will verify if the requested .ko file(s) are present on the host,
   remove from the request missing files and will proceed with the
   removal scenario for the rest
3) unit-test

### DIFF
--- a/internal/utils/filesystem_helper.go
+++ b/internal/utils/filesystem_helper.go
@@ -2,8 +2,10 @@ package utils
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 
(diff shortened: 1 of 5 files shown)


yevgeny-shnaidman:
Fixes[1208](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1208)

PR #1211, opened 2024-09-12
Adapt Day1 feature to support init container in worker pod

TomerNewman:
### PR Details
Today, KMM has an init container in the worker pod. This means that the init container copies all of the .ko files to the shared volume, which will be loaded into the kernel by the worker container.
This commit adds this flow to the Day1 feature.

---

#### Note
I've loaded a simple module into the kernel in a test cluster.

/assign @yevgeny-shnaidman
/cc @ybettan


pkg/mcproducer/scripts/pull-image.sh
```diff
@@ -1,15 +1,13 @@
...
-kernel_module_image_filepath="$KERNEL_MODULE_IMAGE_FILEPATH"
 worker_image="$WORKER_IMAGE"
 kernel_module_image="$KERNEL_MODULE_IMAGE"
 kernel_module_image_tag=$(uname -r)
 full_kernel_module_image="$kernel_module_image:$kernel_module_image_tag"
 
-if [ -e $kernel_module_image_filepath ]; then
-    echo "File $kernel_module_image_filepath found. Nothing to do, the file was handled, removing $kernel_module_image_filepath and $kmm_config_file_filepath"
-    rm -f $kernel_module_image_filepath
+if [ -n "$(podman images -q $worker_image 2> /dev/null)" ]; then
```
yevgeny-shnaidman, line 9: i think you should check the kernel_module_image here, since ti it the second one pulled. in case you check the worker image, and the scripts fails on kernel_module image, then on the next reboot the kernel_module image will never be pulled

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,22 @@ EOF
     cat "$kmm_config_file_filepath"
 }
 
-echo "before checking image tar file presence"
-if [ -e $kernel_module_image_filepath ]; then
-    echo "Image file $kernel_module_image_filepath found on the local file system, creating kmm config file"
+echo "before checking image presence"
+if [ -n "$(podman images -q $worker_image 2> /dev/null)" ]; then
```
yevgeny-shnaidman, line 25: the same here - we should check the second image to be pulled by the pull script

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,22 @@ EOF
...
+    echo "Image $worker_image found on the local file system, creating kmm config file"
     create_kmm_config
-    echo "running kernel-management worker image"
-    podman run --user=root --privileged -v /lib/modules:/lib/modules -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml -v $kernel_module_image_filepath:$kernel_module_image_filepath $worker_image kmod load --tarball /etc/kmm-worker/config.yaml
+    podman pod create --name kmm-pod
+    echo "running init container image"
+    init_container_id=$(podman create --pod kmm-pod --user=root --init-ctr=once --privileged -v /tmp:/tmp $full_kernel_module_image /bin/sh -c "mkdir -p /tmp/opt/lib/modules; cp -R /opt/lib/modules/* /tmp/opt/lib/modules;")
+    podman start "$init_container_id"
```
yevgeny-shnaidman, line 31: question: do we need to start each container in the pod, or can we just start the pod, and it will know how to deal with its containers?

TomerNewman: Thought about it, the only solution I found is to run a YAML.
for example:
```
podman play kube <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
spec:
  initContainers:
  - name: init-container
    image: init-contaier-image
  containers:
  - name: container
    image: container-image
EOF
```
But I thought it will be less clean in the code itself

TomerNewman:
retest

pkg/mcproducer/scripts/replace-kernel-module.sh
yevgeny-shnaidman: I am not sure i understand why we need "podman play kube". You have just created a pod and added 2 containers to ut. Can't we just run "podman pod start <pod name>"?

pkg/mcproducer/scripts/pull-image.sh
```diff
@@ -1,15 +1,13 @@
...
 worker_image="$WORKER_IMAGE"
 kernel_module_image="$KERNEL_MODULE_IMAGE"
 kernel_module_image_tag=$(uname -r)
 full_kernel_module_image="$kernel_module_image:$kernel_module_image_tag"
 
-if [ -e $kernel_module_image_filepath ]; then
-    echo "File $kernel_module_image_filepath found. Nothing to do, the file was handled, removing $kernel_module_image_filepath and $kmm_config_file_filepath"
-    rm -f $kernel_module_image_filepath
+if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
+    echo "Image $worker_image exist locally. Nothing to do, removing $kmm_config_file_filepath"
```
ybettan, line 10:
```suggestion
    echo "Image $full_kernel_module_image exist locally. Nothing to do, removing $kmm_config_file_filepath"
```

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,22 @@ EOF
     cat "$kmm_config_file_filepath"
 }
 
-echo "before checking image tar file presence"
-if [ -e $kernel_module_image_filepath ]; then
-    echo "Image file $kernel_module_image_filepath found on the local file system, creating kmm config file"
+echo "before checking image presence"
+if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
+    echo "Image $worker_image found on the local file system, creating kmm config file"
```
ybettan, line 26:
```suggestion
    echo "Image $full_kernel_module_image found on the local file system, creating kmm config file"
```

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,22 @@ EOF
...
     create_kmm_config
-    echo "running kernel-management worker image"
-    podman run --user=root --privileged -v /lib/modules:/lib/modules -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml -v $kernel_module_image_filepath:$kernel_module_image_filepath $worker_image kmod load --tarball /etc/kmm-worker/config.yaml
+    podman pod create --name kmm-pod
+    echo "running init container image"
+    init_container_id=$(podman create --pod kmm-pod --user=root --init-ctr=once --privileged -v /tmp:/tmp $full_kernel_module_image /bin/sh -c "mkdir -p /tmp/opt/lib/modules; cp -R /opt/lib/modules/* /tmp/opt/lib/modules;")
```
ybettan, line 30: Don't we also need to copy firmware files?
Why is this container has to be `privileged`?

pkg/mcproducer/scripts/replace-kernel-module.sh
ybettan: So `podman` does support running a pod with an init-container eventually?
I definitely prefer applying a yaml - it will be much more similar to what we have in day2.

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,22 @@ EOF
...
+    worker_container_id=$(podman create --pod kmm-pod --user=root --privileged -v /tmp/opt/lib/modules:/tmp/opt/lib/modules:Z -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml $worker_image kmod load --tarball /etc/kmm-worker/config.yaml)
+    podman start "$worker_container_id"
     if [ $? -eq 0 ]; then
         echo "OOT kernel module $kernel_module is inserted"
     else
         echo "failed to insert OOT kernel module $kernel_module"
     fi
 else
-    echo "Image file $kernel_module_image_filepath is not present in local registry, will try after reboot"
+    echo "Image $worker_image is not present in local registry, will try after reboot"
```
ybettan, line 41:
```suggestion
    echo "Image $full_kernel_module_image is not present in local registry, will try after reboot"
```

pkg/mcproducer/scripts/replace-kernel-module.sh
yevgeny-shnaidman: we don't apply a yaml in day2: we simply create a pod with init-container and worker and start it

pkg/mcproducer/scripts/replace-kernel-module.sh
yevgeny-shnaidman: @ybettan why should we copy firmware files? they are part of the container image and will be dealt by init-container

TomerNewman: @ybettan are you sure?
because if I remove` --privileged` I get an error of `mkdir: cannot create directory '/tmp/opt': Permission denied`

pkg/mcproducer/scripts/replace-kernel-module.sh
ybettan: But if we can apply a yaml using `podman` won't it be cleaner? Instead of having a long podman command mounting everything in the same command?

pkg/mcproducer/scripts/replace-kernel-module.sh
ybettan: > they are part of the container image and will be dealt by init-container

@yevgeny-shnaidman the init-container won't do anything with them unless it is told to like @TomerNewman did with the modules files when overriding the entrypoint of the container.

Am I missing something?

ybettan: > @ybettan are you sure?
because if I remove --privileged I get an error of mkdir: cannot create directory '/tmp/opt': Permission denied

I am not sure, but in day 2 only the worker container is privileged - https://github.com/kubernetes-sigs/kernel-module-management/blob/0e806c0d0757e178a001ec8d62e912f5fcdd0a0a/internal/controllers/nmc_reconciler.go#L1144.

yevgeny-shnaidman: @ybettan yes, you are right, i have missed that part

TomerNewman: @ybettan regarding the `--privileged` flag, I removed it.
and following our slack conversation, we will copy the firmware files in the next pr.

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,23 @@ EOF
...
-echo "before checking image tar file presence"
-if [ -e $kernel_module_image_filepath ]; then
-    echo "Image file $kernel_module_image_filepath found on the local file system, creating kmm config file"
+echo "before checking image presence"
+if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
+    echo "Image $full_kernel_module_image found on the local file system, creating kmm config file"
     create_kmm_config
-    echo "running kernel-management worker image"
-    podman run --user=root --privileged -v /lib/modules:/lib/modules -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml -v $kernel_module_image_filepath:$kernel_module_image_filepath $worker_image kmod load --tarball /etc/kmm-worker/config.yaml
+    mkdir /tmp/opt
```
yevgeny-shnaidman, line 28: lets add a rm --rf for that directory, and clean after ourselves

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,23 @@ EOF
...
+    podman pod create --name kmm-pod
+    echo "creating init container"
+    podman create --pod kmm-pod --user=root --init-ctr=always -v /tmp/opt:/tmp/opt:Z $full_kernel_module_image /bin/sh -c "mkdir -p /tmp/opt/lib/modules; cp -R /opt/lib/modules/* /tmp/opt/lib/modules;"
```
ybettan, line 31:
```suggestion
    podman create --pod kmm-pod --user=root --init-ctr=always -v /tmp/opt:/tmp/opt $full_kernel_module_image /bin/sh -c "mkdir -p /tmp/opt/lib/modules; cp -R /opt/lib/modules/* /tmp/opt/lib/modules;"
```

What does that `Z` do?

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,23 @@ EOF
...
+    echo "creating worker container"
+    podman create --pod kmm-pod --user=root --privileged -v /tmp/opt:/tmp/opt -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml $worker_image kmod load --tarball /etc/kmm-worker/config.yaml
```
ybettan, line 33: Let's break this command into multiple lines for clarity?

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,36 @@ EOF
...
+      /bin/sh -c "cp -R /opt/lib/modules/* /tmp/opt/lib/modules"
+    echo "creating worker container"
+    podman create \
+      --pod kmm-pod \
+      --user=root \
+      --privileged \
+      -v /tmp/opt:/tmp/opt \
+      -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml \
+      $worker_image \
+      kmod load --tarball /etc/kmm-worker/config.yaml
```
yevgeny-shnaidman, line 45: we don't need tarball flag anymore

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,36 @@ EOF
...
-echo "before checking image tar file presence"
-if [ -e $kernel_module_image_filepath ]; then
-    echo "Image file $kernel_module_image_filepath found on the local file system, creating kmm config file"
+echo "before checking image presence"
+if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
+    echo "Image $full_kernel_module_image found on the local file system, creating kmm config file"
     create_kmm_config
-    echo "running kernel-management worker image"
-    podman run --user=root --privileged -v /lib/modules:/lib/modules -v $kmm_config_file_filepath:/etc/kmm-worker/config.yaml -v $kernel_module_image_filepath:$kernel_module_image_filepath $worker_image kmod load --tarball /etc/kmm-worker/config.yaml
+    mkdir -p /tmp/opt/lib/modules
```
ybettan, line 28: I would use `podman volume create` instead of manually creating a directory on the host.
https://docs.podman.io/en/latest/markdown/podman-volume-create.1.html

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,36 @@ EOF
...
+    podman pod create --name kmm-pod
+    echo "creating init container"
+    podman create \
+      --pod kmm-pod \
+      --init-ctr=always \
+      -v /tmp/opt/lib/modules:/tmp/opt/lib/modules:Z \
```
yevgeny-shnaidman, line 34: the shared volume between init-container and worker container should be the same on the host. So if you define that the shared volume on the host is /tmp/opt (i would rename opt to something else in order not to confuse with the opt in the container image), than worker container should also mount it

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,36 @@ EOF
...
+      $full_kernel_module_image \
+      /bin/sh -c "cp -R /opt/lib/modules/* /tmp/opt/lib/modules"
+    echo "creating worker container"
+    podman create \
```
yevgeny-shnaidman, line 38: worker container should also mount the /lib/modules host directory

TomerNewman: are you sure? even without this mount I was able to install the kmod on my node

pkg/mcproducer/scripts/replace-kernel-module.sh
ybettan: The day2 solution is mounting an `EmptyDir` (regardless of where it is stored on the host) and mounting it to `/tmp/` in both, the init-container and the worker-container.

When we add firmware, we won't need to add another mount cause they will be looked up somewhere in `/tmp/` as well.

pkg/mcproducer/scripts/replace-kernel-module.sh
yevgeny-shnaidman: yes. It is needed in case the container kernel driver is dependent on the host kernel drivers. In your test case, simple-kmod is not dependent on anything, so it is working

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +21,36 @@ EOF
...
+      $worker_image \
+      kmod load --tarball /etc/kmm-worker/config.yaml
+    echo "running worker pod"
+    podman pod start kmm-pod
     if [ $? -eq 0 ]; then
         echo "OOT kernel module $kernel_module is inserted"
     else
         echo "failed to insert OOT kernel module $kernel_module"
     fi
+    rm -rf /tmp/opt
```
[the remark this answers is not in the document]
TomerNewman: np, just telling it will also delete the containers of the pod (after they are finished to run)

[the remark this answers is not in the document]
ybettan: > np, just telling it will also delete the containers of the pod (after they are finished to run)

I am not sure it should do so. According to the docs only non used volumes will be removed but I don't see why it will also remove the container itself.

```
$ podman volume remove --help
Remove one or more volumes

Description:
  Remove one or more existing volumes.

  By default only volumes that are not being used by any containers will be removed. To remove the volumes anyways, use the --force flag.
```

Am I missing something?

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -19,17 +23,42 @@ EOF
...
+    echo "running worker pod"
+    podman pod start $worker_pod_name
     if [ $? -eq 0 ]; then
         echo "OOT kernel module $kernel_module is inserted"
     else
         echo "failed to insert OOT kernel module $kernel_module"
     fi
+    podman wait $worker_pod_id
+    echo "removing volume"
+    podman volume rm --force $worker_volume_name
```
ybettan, line 61: Why do we need the `--force` flag?

TomerNewman: > Why do we need the `--force` flag?

Because even though the pod is finished, the volume is still being used by the containers. the `--force` will also delete them.

pkg/mcproducer/scripts/replace-kernel-module.sh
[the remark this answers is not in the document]
TomerNewman: If we use podman volume remove <volume_name> it will fail as long as the containers exists even though it was finished to run, thats why I used the `--force` flag

pkg/mcproducer/scripts/replace-kernel-module.sh
ybettan: I think that if we create the containers with `podman create --rm ...` then when the container is done, the container is removed and then we will be able to safely remove the volume.

If it doesn't work, then we can just merge it with the `--force` flag as you did.

WDYT?

TomerNewman: yes sure, it works

PR #1216, opened 2024-09-22
Copying firmware files in day1 feature

TomerNewman:
After making KMM's day1 feature to copy the files using an init container, we also need to support copying firmware files.
This PR adds the option to the user to add the path to the firmware files, and copy them.

---
In addition, this PR deleting the pod in the end of the script.

---


docs/mkdocs/documentation/day1_limited_option.md
```diff
@@ -70,6 +70,8 @@ The parameters are:
                           module
 - `workerImage`: optional parameter. The worker image to use. In case this parameter is not passed, the default worker image 
                  will be used: quay.io/edge-infrastructure/kernel-module-management-worker:latest.
+- `firmwareFilesPath`:` optional parameter. In case there is a need to also use firmware,
+                        this parameter should hold the path to those files as a string format.
```
ybettan, line 74:
```suggestion
                        this parameter should hold the path to the directory containing those files as a string format.
```

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -31,14 +32,27 @@ if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
...
+            --pod $worker_pod_name \
+            --init-ctr=always \
+            --rm \
+            -v $worker_volume_name:/tmp \
+            $full_kernel_module_image \
+            /bin/sh -c "mkdir -p /tmp/opt/lib/modules && \
+            cp -R /opt/lib/modules/* /tmp/opt/lib/modules && \
+            mkdir -p /tmp$firmware_files_path && \
+            cp -R $firmware_files_path/* /tmp$firmware_files_path"
+    fi
```
ybettan, line 55:
```suggestion
    copycmd='mkdir -p /tmp/opt/lib/modules && cp -R /opt/lib/modules/* /tmp/opt/lib/modules'
    if [[ -n "$FIRMWARE_FILES_PATH" ]]; then
      copycmd+=' mkdir -p /tmp/${firmware_files_path} && cp -R ${firmware_files_path}/* /tmp/${firmware_files_path}'
    fi
    podman create \
          --pod $worker_pod_name \
          --init-ctr=always \
          --rm \
          -v $worker_volume_name:/tmp \
          $full_kernel_module_image \
          /bin/sh -c "${copycmd}"
    ```

Comment (ybettan) [reply pkg/mcproducer/scripts/replace-kernel-module.sh:55]: WDYT about this suggestion?

The changes are:
1. Only add the firmeware files in the `if` statement instead of duplicating the `podman command`
2. Split the command using `;` instead of `&&` because there is no reason to fail the second command just because the first one failed
3. Fixed the directory from `/tmp$firmware_files_path` to `/tmp/$firmware_files_path` - how did it work before the fix?
4. Enclosed the variables with `${}` just because it is safer - there is no real use in it here :)

File: pkg/mcproducer/scripts/replace-kernel-module.sh [hunk 3]
```diff
@@ -59,6 +73,8 @@ if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
         echo "failed to insert OOT kernel module $kernel_module"
     fi
     podman wait $worker_pod_id
+    echo "removing kmm-pod"
+    podman pod rm $worker_pod_name
```
ybettan, line 77: Won't the pod be automatically removed if all its containers are using the `--rm` flag?

pkg/mcproducer/testdata/machineconfig-test.yaml
```diff
@@ -21,6 +21,7 @@ spec:
             Type=oneshot
             TimeoutSec=10
             EnvironmentFile=/etc/mco/proxy.env
+            Environment="FIRMWARE_FILES_PATH=/tmp/test"
```
ybettan, line 24: I would use a path that is not in the `/tmp` directory. 
Is this working with `/tmp`? I think that everything in `/tmp` in the container is going to be overwritten by the mounted directory isn't it?

pkg/mcproducer/scripts/replace-kernel-module.sh
TomerNewman: the pod remains with status of `Exited`, even though the containers were removed

pkg/mcproducer/testdata/machineconfig-test.yaml
TomerNewman: The unit test will pass but you're right, I will change it to a different path to make it consistent.

pkg/mcproducer/scripts/replace-kernel-module.sh
TomerNewman: 1. Good idea, it looks much cleaner.
2. Understood, I've added `;` to the end of the copycmd `copycmd="mkdir -p /tmp/opt/lib/modules && cp -R /opt/lib/modules/* /tmp/opt/lib/modules;"`
3. For example lets say the firmware files are in dir `/opt/lib/firmware`, so with
 `/tmp$firmware_files_path` it will be `/tmp/opt/lib/firmware`.
 4. Good idea, also I changed `'` to `"`.

ybettan: > For example lets say the firmware files are in dir /opt/lib/firmware, so with
/tmp$firmware_files_path it will be /tmp/opt/lib/firmware.

This is correct, in KMM's code we use `filepath.Join` to always return a clean path - we need to find a bash solution for it.

pkg/mcproducer/scripts/replace-kernel-module.sh
```diff
@@ -31,14 +32,17 @@ if [ -n "$(podman images -q $full_kernel_module_image 2> /dev/null)" ]; then
     podman volume create $worker_volume_name
     podman pod create --name $worker_pod_name
     echo "creating init container"
+    copycmd="mkdir -p /tmp/opt/lib/modules && cp -R /opt/lib/modules/* /tmp/opt/lib/modules;"
+    if [[ -n "$FIRMWARE_FILES_PATH" ]]; then
+      copycmd+=" mkdir -p /tmp${firmware_files_path} && cp -R ${firmware_files_path}/* /tmp${firmware_files_path}"
```
ybettan, line 37: This won't work if the firmware path doesn't start with `/`

ybettan: Maybe the accepted solution here can be of help
https://unix.stackexchange.com/questions/23208/building-paths-robustly

TomerNewman: Based on the accepted solution and on what we do in kmm, I updated it to be:
`    if [[ -n "$FIRMWARE_FILES_PATH" ]]; then
      folders=("tmp" "$firmware_files_path");
      path_to_copy_firmware=$(printf '/%s' "${folders[@]%/}")
      copycmd+=" mkdir -p ${path_to_copy_firmware} && \
      cp -R ${firmware_files_path}/* ${path_to_copy_firmware}"
    fi
    `
    In kmm we also use filepath.Join function to `*`, but here I think this approach would be more clean this way.

PR #1224, opened 2024-10-07
Fixing yet another race condition during cluster/kernel upgrade

yevgeny-shnaidman:
The following scenario causes worker pods to be stuck during cluster upgrade:
1) kernel module is loaded into the node. NMC contains both spec and status using the current kernel version 2) cluster upgrade starts. As part of the upgrade node becomes
   Unschedulable
3) Module-NMC removes Spec from NMC, since the node is Uschedulable 4) NMC controller creates an unload worker pod, since the timestamp of
   the ready condition of the node has not been updated yet to a new
   value (the node is still not ready).
5) once the node becomes ready, the unload worker pod is scheduled with
   the old kernel version and therefore constantly fails.

Solution:
We will skip handling spec, statuses , garbage collection and labelling for the node that are not ready (unschedulable). We will only collect the completed pod and will update NMC statuses accordingly. Once the node becomes ready, the new reconciliation loop will kick in

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -70,8 +70,9 @@ const (
 //+kubebuilder:rbac:groups="core",resources=serviceaccounts,verbs=get;list;watch
 
 type NMCReconciler struct {
-	client client.Client
-	helper nmcReconcilerHelper
+	client  client.Client
+	helper  nmcReconcilerHelper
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
Fixes [1223](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1223)

PR #1229, opened 2024-10-14
Enhancement for redefining communication between Module-NMC and NMC

yevgeny-shnaidman:
controllers


ybettan:
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1228

PR #1239, opened 2024-11-10
Fixing handling ImageRepoSecret in worker pod

yevgeny-shnaidman:
Since moving to using container-runtime and init-container for pull kernel module container image, we need to set the ImageRepoSecret(if defined) in the worker Pod and also there is no need for mapping ImageRepoSecret and secrets of the SA into worker pod volumes (was need for crane) This PR does the following:
1) set ImageRepoSecret into the worker Pod, if defined in the KMM Module 2) remove creating volumes for SA's secret and ImaRepo secret in the
   worker pod
3) remove pullSecretHelper interface implementation as not needed 4) uni-test updates


yevgeny-shnaidman:
cherry-pick release-2.2

PR #1247, opened 2024-11-14
Add support for user defined tolerances to modules

tsprasannaa:
There is a requirement to run selected pods on a cordoned node. One such example is device driver upgrade. The procedure involves the node to be cordoned. At the same time, there should be a way to run house keeping pods that carry out the driver upgrades. These pods should be schedulable.
This PR adds support to carry the tolerances for the pods that carry out the house keeping operations. These user defined tolerances will match the taint that gets added to the nodes.

ModuleSpec will be used to carry the tolerance to the pods which will be used during pod creation


ybettan:
/hold

@tsprasannaa Thank you for your PR. This is a good improvement. I am OK with the concept of the change but we should apply it u/s first.

If you wish to do it, you can create the same PR at https://github.com/kubernetes-sigs/kernel-module-management and we will review it there, then it will be pulled to this repo.

If you don't want to do it, someone from the team will do it and add you as a co-author to the PR.

Feel free to reach me or @yevgeny-shnaidman on slack if you have any further questions :)

ybettan:
/close
Closing in favor of https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1257

PR #1268, opened 2024-12-12
Making ImagePullPolicy in NMC optional

yevgeny-shnaidman:
Currently ImagePullPolicy has a kubebuilder instruction for default value, which automatically makes it a required field. In previous version, this field was optional, meaning: when upgrading KMM operator, the upgrade will fail , since NMC object will be missing the required field
This commit keep the default value, but makes this field optional, in order to solve the upgrade issue


yevgeny-shnaidman:
cherry-pick release-2.2

PR #1281, opened 2024-12-23
Renaming the webhook deployment from `webhook-server` to `webhook`.

ybettan:
* When we generate files with `controller-gen` it generates a service called `webhook-service` and it is not configurable.

* When we deploy KMM with OLM, OLM deploys a service for the webhook called `<webhook-deployment-name>-service.

Therefore, before this change, we were getting 2 services for the same deployment. One generated by `controller-gen` and added to the bundle manifests and the other one created "on the fly" by OLM.

This change will make OLM find an already existing service called `webhook-service` in the cluster because the deployment is called `webhook`, therefore, it won't create a second service as before.

Services in KMM's namespace before the changes:
```
kmm-operator-controller-metrics-service   ClusterIP   10.129.169.25   <none>  8443/TCP   2d2h
kmm-operator-webhook-server-service       ClusterIP   10.130.107.158  <none>  443/TCP    2d2h
kmm-operator-webhook-service              ClusterIP   10.128.244.171  <none>  443/TCP    2d2h
```

Services in KMM's namespace after the changes:
```
kmm-operator-controller-metrics-service   ClusterIP   10.99.75.82     <none>  8443/TCP   88m
kmm-operator-webhook-service              ClusterIP   10.105.163.241  <none>  443/TCP    88m
```

---

/assign @yevgeny-shnaidman @TomerNewman 
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1280

### DIFF
--- a/bundle-hub/manifests/kernel-module-management-hub.clusterserviceversion.yaml
+++ b/bundle-hub/manifests/kernel-module-management-hub.clusterserviceversion.yaml
@@ -37,7 +37,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2024-12-15T08:56:31Z"
+    createdAt: "2024-12-23T09:43:32Z"
     operatorframework.io/suggested-namespace: openshift-kmm-hub
     operators.operatorframework.io/builder: operator-sdk-v1.32.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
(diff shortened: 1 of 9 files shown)


ybettan:
/retest

On of the services in the Openshift CI cluster is down: https://redhat-internal.slack.com/archives/CBN38N3MW/p1735044015616539

PR #1315, opened 2025-01-28
adding ModuleImagesConfig definition

ybettan:
ModuleImagesConfig CR will be used by the KMM operator to execute image existance flow via kubelet. The CR will hold the image to be verified in the spec, and the state of that image( exists/notexists) in the status field

---

/assign @yevgeny-shnaidman @TomerNewman 
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1300

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Custom Resource Definition (CRD) for managing module image configurations
	- Added support for tracking image states and generations
	- Implemented deep copy methods for new resource types

- **Documentation**
	- Added type definitions and specifications for module image management in Kubernetes


ybettan:
> lets wait until the toleration fixes

Why so? This is the `main` branch. It should not bother us from cherry-picking the toleration commits to the `release-2.2` branch.

PR #1316, opened 2025-01-29
Fixing toleration implementation

yevgeny-shnaidman:
In current implementation, tolerations are stored in the ModuleConfig of the NMC. That means that on each update of Tolerations, kernel module will be unloaded/loaded, which is undesirable.
In this PR, the tolerations are moved into ModuleItem, along with Name, Namespace,Service Account and PullSecret, since it does not affect the kernel module code, but just the way the worker pod is configured to run. This way, tolerations can be changed in whatever way possible, without affecting the loaded kernel modules

## Release Notes

- **New Features**
	- Enhanced module configuration to support more granular toleration management for load and unload pods

- **Changes**
	- Moved tolerations from module configuration to individual module items
	- Updated node module status to reflect pod tolerations
	- Improved CustomResourceDefinition to clarify toleration specifications

- **Technical Improvements**
	- Refined reconciliation logic for handling pod tolerations
	- Updated deep copy mechanisms for module configurations


ybettan:
/hold

We first need to merge other PRs by `gitstream` to make the best effort to have the same history as the u/s repo.

ybettan:
/unhold
Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1328

PR #1257, opened 2024-12-03
Add support for user defined tolerances to modules

TomerNewman:
There is a requirement to run selected pods on a cordoned node. One such example is device driver upgrade. The procedure involves the node to be cordoned. At the same time, there should be a way to run house keeping pods that carry out the driver upgrades. These pods should be schedulable.
This PR adds support to carry the tolerances for the pods that carry out the house keeping operations. These user defined tolerances will match the taint that gets added to the nodes.

ModuleSpec will be used to carry the tolerance to the pods which will be used during pod creation.

---
/cc @yevgeny-shnaidman @ybettan 
fixes #1256 
U/s [PR](https://github.com/kubernetes-sigs/kernel-module-management/pull/940)

PR #1318, opened 2025-01-29
[release-2.3] Add support for user defined tolerances to modules

TomerNewman:
There is a requirement to run selected pods on a cordoned node.
One such example is device driver upgrade. The procedure involves
the node to be cordoned. At the same time, there should be a way
to run house keeping pods that carry out the driver upgrades. These
pods should be schedulable.
This PR adds support to carry the tolerances for the pods that carry
out the house keeping operations. These user defined tolerances will
match the taint that gets added to the nodes.

ModuleSpec will be used to carry the tolerance to the pods which will
be used during pod creation.

---

### Note
Committing also #1253 because it is a dependency of #1257

---


ybettan:
/hold

I will create the test file to run CI on the `release-2.3` branch.

Also, we should rename the commit message to remove the [Cherrypick-release2.3 - ] prefix. The commit message doesn't care to what branch it is applied.

We can keep it in the PR description only as `[relelase-2.3]` because this is what the openshift bot does and it is easy to understand when reviewing.

ybettan:
Once https://github.com/openshift/release/pull/61081 merged, CI should run on this PR (we may need to re-push to trigger).

ybettan:
This PR is missing a "father commit" to work correctly. It will be re-created with 2 commits in it.

TomerNewman:
/unhold
/override check-commits-count
This PR will hold 2 commits, the desired toleration support commit and his dependency commit.

TomerNewman:
/override ci/prow/security
same as Yoni mentioned, 
```
Unsanitized input from a CLI argument flows into os.exec.Command, where it is used as a shell command. This may result in a Command Injection vulnerability.
‎internal/worker/cmdlog_test.go:58
```

TomerNewman:
/override ci/prow/security /override ci/prow/check-commits-count

ybettan:
CI should be fixed by merging https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1335

PR #1411, opened 2025-03-19
Changing MSBC image status structure

yevgeny-shnaidman:
Currently MBSC image's spec contains the action that MBSC controller should execute(sign or build). But since the action does not exists in the image's status, the MIC controller cannot know if the success status refers to the previous action (i.e build) or the current requested action( i.e sign). In addition, since action will appear in status, there is no need to have explicit statuses like BuildFailed: we change it to ActionFailed or ActionSuccess.
This PR contains the following:
1) adding action field to the MBSC image' status
2) chaning the type of Status field: now it will be only ActionFailure
   or ActionSuccess
2) update the MIC controller MBSC lookup function to decide what is the
   current status of image in MIC, based on MBSC image's action and
status fields
3) unit-tests


yevgeny-shnaidman:
Fixes [1386](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1386)

PR #1416, opened 2025-03-19
Adding RegistryTLS to the MIC image spec

yevgeny-shnaidman:
MIC image spec is being propagated as is to the MBSC image spec. As part of building sign and build pod, one of the inputs is the RegistryTLS configuration used during build/sign process. This PR does the following:
1. Add RegistryTLS field to the ModuleImageSpec
2. module reconciler sets the RegistryTLS in the imageSpec variable
3. changing unit-tests

## Summary by CodeRabbit

- **New Features**
  - Introduced configurable TLS options for registry access in module management. Users can now enable insecure connections and skip TLS verification when needed, providing enhanced flexibility over registry connectivity settings.


yevgeny-shnaidman:
Fixes [1407](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1407)

PR #1421, opened 2025-03-20
Fixing the firmware configuration documentation.

yevgeny-shnaidman:
The config fields to be used in order to modify the firmware path on the host but the user was changed rom `setFirmwareClassPath` to `firmwareHostPath` but the documentation wasn't update accordingly. This commit is about to fix that.

## Summary by CodeRabbit

- **Documentation**
	- Updated the header in the worker configuration documentation to reflect the revised naming convention for the firmware path, while keeping the descriptive details and recommended values unchanged.


yevgeny-shnaidman:
Fixes [1419](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1419)

PR #1439, opened 2025-04-10
Introducing buildsign package

yevgeny-shnaidman:
In the current implementation of the build and sign flow, build_sign controller uses 2 packages, build and sign, to verify whether build/sign must be executed, to actually create build/sign pod and to run garbage collector on he successfull pods. In many cases, those 2 packages are very much the same, especially in where the package manager must decide what to do (build pod, delete pod, etc)
With the introduction of the MBSC controller, the flow is changing and it makes sense to unify those packages into one package. There is not need to decide if the build/sign should be executed, since that decision is taken by the MIC controller

This PR starts with creating the buildsing package and unifying Helper interfaces both in sign and build package into one Helper interface: 1) create Helper interface and implementation for buildsign package 2) move the existing code that was previously using build.Helper and
   sign.Helper interface to start using unified interface


yevgeny-shnaidman:
Fixes [1420](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1420)

PR #1444, opened 2025-04-23
Adding OCP Build manager

yevgeny-shnaidman:
OCP Build Manager is responsible for for accessing the OCP Build objects used by buildsign package both for building and signing flow. This is the porting layer and is the downsteam parallel of the buildSignPodManager

## Summary by CodeRabbit

- **New Features**
  - Introduced management of OpenShift Container Platform (OCP) build resources for kernel modules, including creation, deletion, status tracking, and filtering by module, kernel version, and build type.
  - Added mock implementations to support testing of OCP build management operations.

- **Tests**
  - Added comprehensive unit tests for all OCP build management operations, ensuring robust error handling and correct behavior across different scenarios.
  - Established a new test suite for the OCP build management component.


internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+//go:generate mockgen -source=ocpbuildmanager.go -package=ocpbuild -destination=mock_ocbuildmanager.go
+
+type ocpbuildManager interface {
+	isOCPBuildChanged(existingBuild *buildv1.Build, newBuild *buildv1.Build) (bool, error)
+	getModuleOCPBuildByKernel(ctx context.Context, modName, namespace, kernelVersion, ocbuildType string, owner metav1.Object) (*buildv1.Build, error)
+	getModuleOCPBuilds(ctx context.Context, modName, modNamespace, ocbuildType string, owner metav1.Object) ([]buildv1.Build, error)
+	getOCPBuildStatus(build *buildv1.Build) (Status, error)
+	createOCPBuild(ctx context.Context, build *buildv1.Build) error
+	deleteOCPBuild(ctx context.Context, build *buildv1.Build) error
+	ocpbuildLabels(modName, kernelVersion, ocpbuildType string) map[string]string
```
ybettan, line 40:
```suggestion
	getOCPBuildLabels(modName, kernelVersion, ocpbuildType string) map[string]string
```

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	owner metav1.Object) (*buildv1.Build, error) {
+	buildList := buildv1.BuildList{}
+
+	opts := []client.ListOption{
+		client.MatchingLabels(omi.ocpbuildLabels(modName, kernelVersion, ocbuildType)),
+		client.InNamespace(modNamespace),
+	}
+
+	if err := omi.client.List(ctx, &buildList, opts...); err != nil {
+		return nil, fmt.Errorf("could not list Build: %v", err)
```
ybettan, line 79:
```suggestion
		return nil, fmt.Errorf("could not list builds: %v", err)
```

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// filter OCP builds by owner, since they could have been created by the preflight
+	// when checking that specific module
+	moduleOwnedOCPBuilds := filterOCPBuildsByOwner(buildList.Items, owner)
+
+	if n := len(moduleOwnedOCPBuilds); n == 0 {
+		return nil, ErrNoMatchingBuild
```
ybettan, line 87: Why is it an error? Can we just return nil with no error if no builds exist for the module?

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+func (omi *ocpbuildManagerImpl) getModuleOCPBuilds(ctx context.Context, modName, modNamespace, ocpbuildType string, owner metav1.Object) ([]buildv1.Build, error) {
+	buildList := buildv1.BuildList{}
+
+	opts := []client.ListOption{
+		client.MatchingLabels(moduleLabels(modName, ocpbuildType)),
+		client.InNamespace(modNamespace),
+	}
+
+	if err := omi.client.List(ctx, &buildList, opts...); err != nil {
+		return nil, fmt.Errorf("could not list Build: %v", err)
```
ybettan, line 104:
```suggestion
		return nil, fmt.Errorf("could not list builds: %v", err)
```

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	if n := len(moduleOwnedOCPBuilds); n == 0 {
+		return nil, ErrNoMatchingBuild
+	} else if n > 1 {
+		return nil, fmt.Errorf("expected 0 or 1 Builds, got %d", n)
+	}
+
+	return &moduleOwnedOCPBuilds[0], nil
+}
+
+func (omi *ocpbuildManagerImpl) getModuleOCPBuilds(ctx context.Context, modName, modNamespace, ocpbuildType string, owner metav1.Object) ([]buildv1.Build, error) {
```
ybettan, line 95: This is the same method as the one before excluding 1 kernel label. Can me use the same method with a list of labels as arg or something similar instead of duplicating the code?

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+func (omi *ocpbuildManagerImpl) createOCPBuild(ctx context.Context, build *buildv1.Build) error {
+	err := omi.client.Create(ctx, build)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
ybettan, line 133: Do we need this method? calling it or calling directly `client.Create` will look the same.

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (omi *ocpbuildManagerImpl) deleteOCPBuild(ctx context.Context, build *buildv1.Build) error {
+	opts := []client.DeleteOption{
+		client.PropagationPolicy(metav1.DeletePropagationBackground),
```
ybettan, line 137: What does that do?

internal/buildsign/ocpbuild/ocpbuildmanager.go
yevgeny-shnaidman: it is not a set operation, it  just returns the labels. In the upstream we have a similar function, which is called PodLabels

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+		client.InNamespace(modNamespace),
+	}
+
+	if err := omi.client.List(ctx, &buildList, opts...); err != nil {
+		return nil, fmt.Errorf("could not list Build: %v", err)
+	}
+
+	// filter OCP builds by owner, since they could have been created by the preflight
+	// when checking that specific module
+	moduleOwnedOCPBuilds := filterOCPBuildsByOwner(buildList.Items, owner)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: no, filtering is done based on labels, while owner is completely different animal

internal/buildsign/ocpbuild/ocpbuildmanager.go
yevgeny-shnaidman: we want a specific error that says that the Build does not exists, but that the operation itself have not failed. The same as client returns when you try to get an object and the operation is successful but the object was not found

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if n := len(moduleOwnedOCPBuilds); n == 0 {
+		return nil, ErrNoMatchingBuild
+	} else if n > 1 {
+		return nil, fmt.Errorf("expected 0 or 1 Builds, got %d", n)
+	}
+
+	return &moduleOwnedOCPBuilds[0], nil
```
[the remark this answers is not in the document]
yevgeny-shnaidman: because we are using the List command

internal/buildsign/ocpbuild/ocpbuildmanager.go
yevgeny-shnaidman: fixed

internal/buildsign/ocpbuild/ocpbuildmanager.go
yevgeny-shnaidman: They are small functions and signature is not the same: one returns a pointer to a single Build, while the other returns a list of Builds. There is a common code (List), but moving it to a separate function does not reduce the number of lines dramatically. In upstream we have the same functions also

ybettan:
Are we going to remove `internal/utils/ocpbuild/helper.go` in a later PR?

internal/buildsign/ocpbuild/ocpbuildmanager.go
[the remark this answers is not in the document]
ybettan: My question was why do we don't use a `client.Get` method with a `[]client.GetOption` instead?

yevgeny-shnaidman:
> Are we going to remove `internal/utils/ocpbuild/helper.go` in a later PR?

yes, once we move to the new implementation, that code will be removed

internal/buildsign/ocpbuild/ocpbuildmanager.go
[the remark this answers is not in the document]
yevgeny-shnaidman: you can use get in case you know the exact Name and Namespace of the object. In our case we know the namespace, but not the name (it is generated), so we have to use List and labels

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	case buildv1.BuildPhaseFailed, buildv1.BuildPhaseCancelled, buildv1.BuildPhaseError:
+		return StatusFailed, nil
+	default:
+		return "", fmt.Errorf("unknown status: %v", build.Status.Phase)
+	}
+}
+
+func (omi *ocpbuildManagerImpl) createOCPBuild(ctx context.Context, build *buildv1.Build) error {
+	err := omi.client.Create(ctx, build)
+	if err != nil {
```
TomerNewman, line 129: You can do `return omi.client.Create(ctx, build)` instead, the function will behave the same.

TomerNewman: Oh its merged

PR #1454, opened 2025-04-29
Adding GetImageStatus API to mbsc package

yevgeny-shnaidman:
This API return the status of the an image in the MBSC object based on the image and the action. Is required by the MBSC controller

## Summary by CodeRabbit

- **New Features**
  - Added the ability to retrieve the status of a specific image and action within module build and sign configurations.

- **Tests**
  - Introduced new tests to verify correct status retrieval for various image and action scenarios.

- **Chores**
  - Updated mock interfaces to support the new status retrieval functionality for improved test coverage.


yevgeny-shnaidman:
Fixes [1453](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1453)

PR #1458, opened 2025-04-30
Fixing MBSC object's owner creation

yevgeny-shnaidman:
When MBSC object is created/updated by MIC controler , its owner reference should be set with controller=true ,in order to be able to receive events based on the "Own" controller-runtime construct

## Summary by CodeRabbit

- **Refactor**
  - Updated ownership handling to establish a controller-owner relationship between related configuration objects. No changes to user-facing functionality.


yevgeny-shnaidman:
Fixes [1455](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1455)

PR #1466, opened 2025-05-03
Filtering out node heartbeats events for the NMC controller.

yevgeny-shnaidman:
Node heartbeats are there to let the k8s API server that the node is still connected and functional and if not filtered, it will spam the events the NMC controller gets.

The NMC controller, is trying to garbage collect pods for NMCs that were removed - it causes a constant reconciliation even when no Module, and therefore no NMC, are applied to the cluster.

This commit is fixing this issue.

## Summary by CodeRabbit

- **New Features**
  - Improved filtering of node update events to ignore changes related only to heartbeat timestamps, reducing unnecessary reconciliations.
- **Bug Fixes**
  - Enhanced event handling to better distinguish meaningful node changes from routine status updates.
- **Tests**
  - Added tests to verify correct filtering of node update events based on heartbeat and other significant changes.

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -200,7 +200,7 @@ func (r *NMCReconciler) SetupWithManager(ctx context.Context, mgr manager.Manage
 		Watches(
 			&v1.Node{},
 			handler.EnqueueRequestsFromMapFunc(nodeToNMCMapFunc),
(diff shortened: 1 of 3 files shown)


yevgeny-shnaidman:
Fixes [1463](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1463)

PR #1468, opened 2025-05-04
KMM Module should be Owner-Controller of MIC

yevgeny-shnaidman:
In order to receive event from MIC, KMM Module should be the controller of the MIC, and not just an owner

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved how ownership metadata is set for resources, ensuring the correct controller is assigned.
	- Updated error handling in tests to more accurately reflect errors from resource retrieval operations.


yevgeny-shnaidman:
Fixes [1464](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1464)

PR #1473, opened 2025-05-04
Passing the MBSC object as an input to buildsign.Manager interface

yevgeny-shnaidman:
Manager interface receives the meta.Object as an input parameter. It passes it to SetControllerReference function, which also expects meta.Object. But inside the function, the it is cast into runtime.Object interface, which requires implementation of DeepCopy function, which is implemented by the whole MBSC object

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability by updating how certain internal operations handle object data, ensuring more accurate processing.

- **Tests**
  - Updated tests to align with changes in object handling, maintaining consistency and test coverage.


yevgeny-shnaidman:
Fixes [1467](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1467)

PR #1481, opened 2025-05-07
Start using `MIC` status to check if an image exists.

ybettan:
Now that the MIC object is being reconciled correctly then we can start using the MIC.status in order to check if an image exists instead of using an explicit HTTP request from the controller container.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1479
/assign @yevgeny-shnaidman 

## Summary by CodeRabbit

- **New Features**
  - Improved reliability by verifying container image existence through a new configuration resource rather than direct registry checks.
- **Bug Fixes**
  - Enhanced error handling when retrieving image configuration data.
- **Tests**
  - Updated and expanded tests to reflect the new image existence verification approach and to cover additional error scenarios.


ybettan:
/hold

Should only be merged after https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1482

PR #1484, opened 2025-05-07
Changing ImagePuller interface to more elaborate parameters

yevgeny-shnaidman:
Since ImagePuller interface is going to be used not only by MIC, but also by Preflight controller, the APIs should be changed to receive specific parameters, and not use ModuleImagesConfig or ImageSpec structures, that are alient to Preflight

## Summary by CodeRabbit

- **Refactor**
  - Streamlined pod management logic by generalizing parameters for pod creation and listing, reducing dependencies on specific object types.
  - Updated pod labeling for ownership to improve clarity and flexibility.
- **Tests**
  - Adjusted test cases to align with the new parameter structure and updated label keys for better consistency.
- **Chores**
  - Updated mock interfaces and method signatures to match the refactored implementation.

### DIFF
--- a/internal/controllers/mic_reconciler.go
+++ b/internal/controllers/mic_reconciler.go
@@ -74,7 +74,7 @@ func (r *micReconciler) Reconcile(ctx context.Context, micObj *kmmv1beta1.Module
 		return res, nil
 	}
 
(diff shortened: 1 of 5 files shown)


yevgeny-shnaidman:
Fixes [1478](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1478)

PR #1493, opened 2025-05-14
Refactoring the way we get the hash annotation for build/sign.

ybettan:
Reducing the diff between the 2 flows and re-use code.

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1489
Some cosmetic changes were made to make the code a bit more like the u/s code

/assign @TomerNewman @yevgeny-shnaidman 

## Summary by CodeRabbit

- **Refactor**
	- Improved separation of logic for handling secrets, build filtering, labeling, and hashing, resulting in more modular and maintainable code.
	- Standardized annotation keys across the application for consistency.
- **Tests**
	- Enhanced test reliability by improving secret data mocking and updating tests to use the standardized annotation key.
- **Chores**
	- Removed unused methods and updated internal references to streamline the codebase.
- **New Features**
	- Added a new annotation key for tracking the last hash related to modules or nodes.

### DIFF
--- a/internal/buildsign/ocpbuild/common.go
+++ b/internal/buildsign/ocpbuild/common.go
@@ -9,7 +9,9 @@ import (
 	"github.com/mitchellh/hashstructure/v2"
 	buildv1 "github.com/openshift/api/build/v1"
 	kmmv1beta1 "github.com/rh-ecosystem-edge/kernel-module-management/api/v1beta1"
+	"github.com/rh-ecosystem-edge/kernel-module-management/internal/constants"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 )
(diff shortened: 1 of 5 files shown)


internal/buildsign/ocpbuild/common.go
```diff
@@ -73,44 +75,106 @@ func buildVolumesFromBuildSecrets(secrets []v1.LocalObjectReference) []buildv1.B
...
+
+	sourceConfigHash, err := hashstructure.Hash(sourceConfig, hashstructure.FormatV2, nil)
 	if err != nil {
-		return 0, fmt.Errorf("could not hash build spec and secrets: %v", err)
+		return 0, fmt.Errorf("could not hash Build's Buildsource template: %v", err)
 	}
-	return hashValue, nil
+
+	return sourceConfigHash, nil
 }
```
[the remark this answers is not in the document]
ybettan: This is another effect of what we discussed about @yevgeny-shnaidman

ybettan:
/unhold

We won't change the build annotation right now. We decided to keep the existing behavior for now and open a task to investigate it for future releases.

internal/buildsign/ocpbuild/ocpbuildmanager.go
```diff
@@ -35,8 +32,6 @@ const (
 	StatusCreated    Status = "created"
 	StatusInProgress Status = "in progress"
 	StatusFailed     Status = "failed"
-
-	hashAnnotation = "kmm.node.kubernetes.io/last-hash"
```
yevgeny-shnaidman, line 39: Why do we need to remove it? This is an internal constant of the buildsign package, it is not used anywhere else, why make it public?

ybettan: Just to make it more consistent with the [u/s code](https://github.com/kubernetes-sigs/kernel-module-management/blob/31aa5d0d89d63a809c8e7de83eae547666de285e/internal/constants/constants.go#L8). We can indeed stop exporting it but let's merge this one and fix it u/s first (to then be backported here).

PR #1512, opened 2025-05-21
Adding SkipWaitMissingImage field to MIC Spec

yevgeny-shnaidman:
Currently, in case image is missing and Build and Sign sections are not defined, MIC controller will keep the image puller pod running and waiting till the images appears. This means that pushing the mising image into the repo will cause the image puller to succeed and the KMM flow will continue running without any need for additional user action. The new field, if set to true, will cause the MIC to report the image as missing and abort the image puller pod (in case Build and Sign sections are not defined). Currently it will be used in the Preflight flow

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to allow images to be reported as missing immediately if no build or sign instructions are provided, reducing unnecessary waiting time in relevant configurations.

### DIFF
--- a/api/v1beta1/moduleimagesconfig_types.go
+++ b/api/v1beta1/moduleimagesconfig_types.go
@@ -50,6 +50,10 @@ type ModuleImageSpec struct {
 	// +optional
 	Sign *Sign `json:"sign,omitempty"`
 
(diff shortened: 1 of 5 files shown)


yevgeny-shnaidman:
Fixes [1510](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1510)

PR #1529, opened 2025-05-22
Change Preflight VerifactionStatus and VerifactionStage definitions

yevgeny-shnaidman:
Since preflight is moving to MIC/MBSC implementation, the preflight implementation is also changing. MIC is taking over all the handling of build/sign and image verification, so we only need one stage: Image, and the status is now becoming InProgress/Success/Failure

### DIFF
--- a/api/v1beta1/preflightvalidation_types.go
+++ b/api/v1beta1/preflightvalidation_types.go
@@ -26,13 +26,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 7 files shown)


yevgeny-shnaidman:
Fixes [1528](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1528)

PR #1571, opened 2025-06-09
Updating imagepuller pod to use ImagePullPolicy

yevgeny-shnaidman:
Since imagepuller pod must use the same pull policy as worker pod, this commit updates the imagepuller interface to use the pull policy from the MIC object. This commit also updates the unit tests and the mic controller to pass the pull policy from MIC object to the imagepuller interface


yevgeny-shnaidman:
Fixes [1568](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1568)

PR #1575, opened 2025-06-10
removing duplicate imageSpec from MIC

yevgeny-shnaidman:
CreateOrPatch MIC API receives slice of imageSpecs. Since usually the input is created by going over the the target nodes and adding their images to the slice, we need to remove duplicate images, since for most of the cases all the nodes will have the same target image

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of image lists by automatically removing duplicate images before processing.

- **Tests**
  - Added tests to ensure duplicate images are correctly filtered out and only unique images are retained.


yevgeny-shnaidman:
Fixes [1570](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1570)

PR #1588, opened 2025-06-19
Adding PreflightValidationsOCP controller

yevgeny-shnaidman:
This controller will handle the PreflightValidationsOCP object. The flow is simple:
1. get DTK image from the CR and set the kernelOsDtkMapping with appropriate values
2. create/patch PreflightValidations CR with the kernel version and push flag from the PreflightValidationsOCP CR. this will kick off the PreflightValidations controller flow
3. update PreflightValidationsOCP CR's status withe the status of PreflightValidations CR


internal/controllers/preflightvalidationocp_reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+			Name:      pvo.Name,
+			Namespace: pvo.Namespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrPatch(ctx, porhi.client, &pvObj, func() error {
+		pvObj.Spec.KernelVersion = pvo.Spec.KernelVersion
+		pvObj.Spec.PushBuiltImage = pvo.Spec.PushBuiltImage
+
+		return controllerutil.SetOwnerReference(pvo, &pvObj, porhi.scheme)
```
ybettan, line 147: Didn't we say we need `DetControllerReference` instead?

PR #1606, opened 2025-06-22
Adding PushBuiltImage variable to MIC and MBSC objects

yevgeny-shnaidman:
This flag will be used to signal whether the built object must be also push to the destination registry. The main user will be the Preflight. In the regular flow (Module) the image that is built is also always pushed. When running Preflight, user might specify whether to push the built image, or not (he wants to verify that image can be built in time, but does not want to push it right now)

## Summary by CodeRabbit

- **New Features**
  - Added a new optional setting to module image and build sign configurations to control whether built images should also be pushed to a specified repository.

- **Documentation**
  - Updated CustomResourceDefinition (CRD) schemas to include the new push option for built images.

- **Chores**
  - Updated metadata timestamps in ClusterServiceVersion manifest files.


yevgeny-shnaidman:
Fixes [1602](https://github.com/rh-ecosystem-edge/kernel-module-management/issues/1602)

PR #1641, opened 2025-08-13
Adding Network policies for KMM from the operator code

TomerNewman:
KMM's operands should have network policies and because they are being created dynamically, KMM should create the Network Policies dynamically.

This commit creates the network policies on the Modules namespace.

---


config/network-policy/networkpolicy-controller.yaml
```diff
@@ -0,0 +1,33 @@
...
+            matchLabels:
+              dns.operator.openshift.io/daemonset-dns: default
+      ports:
+        - protocol: UDP # DNS
+          port: 53
+        - protocol: TCP # DNS
+          port: 53
+    - ports: # kube api server
+        - protocol: TCP
+          port: 6443
```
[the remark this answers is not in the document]
TomerNewman: https://kubernetes.io/docs/concepts/security/controlling-access/#transport-security

it can listen to 443 also

config/network-policy/networkpolicy-build-and-sign.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: NetworkPolicy
+metadata:
+  name: networkpolicy-build-and-sign
+  namespace: system
+spec:
+  podSelector:
+    matchExpressions:
+    - key: openshift.io/build.name
+      operator: Exists
+  policyTypes:
```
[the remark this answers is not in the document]
TomerNewman: no its ok, both sign and build are build pods running based on Openshift's Builds

config/network-policy/networkpolicy-webhook.yaml
```diff
@@ -0,0 +1,22 @@
...
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 9443
+  egress:
+  - ports: # kube api server port
+    - protocol: TCP
+      port: 6443
+    - protocol: TCP
+      port: 443
```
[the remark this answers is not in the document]
TomerNewman: it is in config/deploy/kustomization.yaml

config/olm/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ namePrefix: kmm-operator-
 resources:
 - ../deploy
 labels:
-- includeSelectors: true
+- includeSelectors: false
```
ybettan, line 13: I think this file can be completely removed (in another, u/s first, preliminary PR). We don't use the CA from the node anymore - this was part of the old implementation when we pulled images using HTTP.

@yevgeny-shnaidman Do you see any reason to keep this one?

TomerNewman:
/hold

This PR will be draft until we will decide to go with that approach

internal/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+	return &networkPolicy{
+		client: client,
+		scheme: scheme,
+	}
+}
+
+const (
+	workerPodNetworkPolicyName     = "kmm-worker-pods"
+	buildSignPodsNetworkPolicyName = "kmm-build-and-sign"
+	pullPodNetworkPolicyName       = "kmm-pull-pods"
```
ybettan, line 41: I would add `kmm-network-policy-*-pods` for all of the consts here.

internal/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+func (np *networkPolicy) CreateOrAddOwnerReference(ctx context.Context, networkPolicy *networkingv1.NetworkPolicy, owner metav1.Object) error {
+	logger := log.FromContext(ctx)
+	logger.V(1).Info("Creating or patching NetworkPolicy", "name", networkPolicy.Name, "namespace", networkPolicy.Namespace, "owner", owner.GetName())
+
+	existing := &networkingv1.NetworkPolicy{}
+	key := client.ObjectKey{Name: networkPolicy.Name, Namespace: networkPolicy.Namespace}
+
+	err := np.client.Get(ctx, key, existing)
+	if err == nil {
+		return np.ensureOwnerReference(ctx, existing, owner)
```
ybettan, line 53: What does that do? Why do we need it?

internal/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+						Operator: metav1.LabelSelectorOpExists,
+					},
+				},
+			},
+			PolicyTypes: []networkingv1.PolicyType{
+				networkingv1.PolicyTypeEgress,
+			},
+			Egress: []networkingv1.NetworkPolicyEgressRule{
+				{}, // Allow all egress
+			},
```
ybettan, line 136: Builds needs to pull and push aren't they? Don't we need to allow incoming/outgoing networking for those?

internal/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      pod.PullPodTypeLabelKey,
+						Operator: metav1.LabelSelectorOpExists,
+					},
+				},
+			},
+			PolicyTypes: []networkingv1.PolicyType{
+				networkingv1.PolicyTypeIngress,
+				networkingv1.PolicyTypeEgress,
```
ybettan, line 158: Is this a deny-all policy? (I deduced it from the worker policy). We only need incoming traffic isn't it?

internal/networkpolicy/networkpolicy.go
TomerNewman: But if the resource's kind is already network policy, why also to put in the name "network-policy"?
regarding the second part, I agree.

internal/controllers/module_reconciler.go
```diff
@@ -535,6 +549,32 @@ func (mrh *moduleReconcilerHelper) moduleUpdateWorkerPodsStatus(ctx context.Cont
...
+
+	if err := mrh.networkPolicyAPI.CreateOrAddOwnerReference(ctx, buildSignPolicy, moduleOwner); err != nil {
+		return fmt.Errorf("failed to ensure BuildSignPods NetworkPolicy %s/%s: %v", buildSignPolicy.Namespace, buildSignPolicy.Name, err)
+	}
+
+	if err := mrh.networkPolicyAPI.CreateOrAddOwnerReference(ctx, pullPodPolicy, moduleOwner); err != nil {
+		return fmt.Errorf("failed to ensure PullPod NetworkPolicy %s/%s: %v", pullPodPolicy.Namespace, pullPodPolicy.Name, err)
+	}
+
+	logger.Info("Successfully handled NetworkPolicies")
```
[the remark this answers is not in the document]
TomerNewman: After talking about it on Slack, I will comment here for protocol:
NPs does not have status so we can not track in a clean way when they have enforced by the network pluggin.
Another solution I thought about will be to not do anything, if for example Build/Sign pod will try to pull a base image and the rule will not be enforced it, it will retry to do until it succeed.

internal/networkpolicy/networkpolicy.go
ybettan: I think writing it simpler will be cleaner and clearer. My understanding is that currently we allow everything but the way it is written is by:
* Ingress - allow all because it is not listed in the `PolicyTypes`
* Egress - allow all because it is listed in the `PolicyTypes` and the rule in the `Egress` section is "allow all"

So we have 2 different ways of allowing traffic which medially raise questions.

I think writing it as the follow will have the same effect and will be cleaner:
```
func (np *networkPolicy) BuildSignPodsNetworkPolicy(namespace string) *networkingv1.NetworkPolicy {
    return &networkingv1.NetworkPolicy{                                                            
        ObjectMeta: metav1.ObjectMeta{                                                             
            Name:      buildSignPodsNetworkPolicyName,                                             
            Namespace: normalizeNamespace(namespace),
        },
        Spec: networkingv1.NetworkPolicySpec{                                                      
            PodSelector: metav1.LabelSelector{                                                     
                MatchExpressions: []metav1.LabelSelectorRequirement{                               
                    {                                                                              
                        Key:      "openshift.io/build.name",                                       
                        Operator: metav1.LabelSelectorOpExists,                                    
                    },                                                                             
                },                                                                                 
            },
        },
    }
}

```

PR #1652, opened 2025-08-26
Adding network policies to kmm including to kmm bundle

TomerNewman:
Due to security concerns, we need to allow KMM
operator only the nessecery traffic.
This commits adds Network policies for each kmm pod.
1. controller
2. webhook
3. build and sign This commit also changes e2e tests to verify the network policies affect.
This commit also affects KMM bundle to include the network policy mainfests.

---

/hold 

## Release Notes

* **New Features**
  * Added network policies for enhanced network security and isolation for controller and webhook components.

* **Configuration Updates**
  * Simplified pod labeling for cleaner resource identification.
  * Updated service routing configuration to align with new label scheme.
  * Configured namespace targeting for test environments.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,8 +279,8 @@ operator-sdk:
 		chmod +x ${OPERATOR_SDK}; \
 	fi
 
-.PHONY: bundle
-bundle: operator-sdk manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+.PHONY: bundle-old
+bundle-old: operator-sdk manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
 	rm -fr ./bundle
 	${OPERATOR_SDK} generate kustomize manifests --apis-dir api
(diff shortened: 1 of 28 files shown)


ybettan:
We are currently waiting until OLM is stable in older versions of OCP and Konflux allows NP manifests in the bundles.
Estimation is it will be done ~Q1 2026.

bundle/manifests/kmm-operator-build-and-sign_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: build-and-sign
```
ybettan: We already shipped NP for operands, we should now only add NP manifests to the bundle for the operator itself (controllers + webhooks)

PR #1657, opened 2025-09-05
Introducing mcfg (machine config) package

yevgeny-shnaidman:
Currently pkg/mcproduce includes a code that creates the MachineConfig for Day1. we are going to have a controller that will also need to create/update a MachineConfig, specifically ignition part. This commit introduces a mcfg package, that contains API that creates an ignition json, and will be used by the mcproducer and the future controller. Changes:
1. templates and scripts involved in the ignition are moved to mcfg package
2. an API for generating the Ignition (using scripts and templates) is introduced
3. in pkg/mcproduces, the machine config package is changed to just imprt the ignition created by the mcfg API

## Summary by CodeRabbit

* **New Features**
  * MachineConfigs now include a generated Ignition payload that provisions and sequences services and helper scripts (crio-wipe remains masked).

* **Refactor**
  * Templates simplified to accept a pre-rendered Ignition blob; ignition generation moved out of template rendering.

* **Tests**
  * New test suite verifies generated Ignition YAML/JSON matches expected fixtures.

* **Chores**
  * Updated dependencies and vendored libraries (adds template helpers, semver, string utilities, decimal, crypto helpers).


internal/mcfg/testdata/ignition-test.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - contents: |
+            [Unit]
+            Description=Replace in-tree kernel module with oot kernel module
+            Before=network-pre.target
+            Wants=network-pre.target
+            DefaultDependencies=no
+            [Service]
+            User=root
+            Type=oneshot
+            TimeoutSec=10
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

internal/mcfg/mcfg.go
```diff
@@ -0,0 +1,82 @@
...
+	templateFS embed.FS
+
+	ignitionTemplate = template.Must(
+		template.ParseFS(templateFS, "templates/ignition.gotmpl"),
+	)
+)
+
+//go:generate mockgen -source=mcfg.go -package=mcfg -destination=mock_mcfg.go
+
+type MCFG interface {
```
ybettan, line 38:
```suggestion
type MC interface {
```

MC is the common acronym for MachineConfig in OCP.

internal/mcfg/mcfg.go
```diff
@@ -0,0 +1,82 @@
...
+	GenerateIgnition(kernelModuleImage, kernelModuleName, inTreeModuleToRemove, firmwareFilesPath, workerImage, servicePrefix string) ([]byte, error)
+}
+
+type mcfgImpl struct {
```
ybettan, line 42:
```suggestion
type mc struct {
```
WDYT?

yevgeny-shnaidman: The same answer as above

internal/mcfg/mcfg.go
```diff
@@ -0,0 +1,81 @@
...
+		"WaitForNetworkDispatcherContents": base64.StdEncoding.EncodeToString([]byte(scriptWaitForNetworkDispatcher)),
+	}
+
+	var parsedIgnition bytes.Buffer
+
+	if err := ignitionTemplate.Execute(&parsedIgnition, templateParams); err != nil {
+		return nil, fmt.Errorf("could not render the ignition: %v", err)
+	}
+
+	var yamlIgnition map[string]interface{}
```
ybettan, line 70: Can't we json-marshal directly from a map?

internal/mcfg/templates/ignition.gotmpl
```diff
@@ -0,0 +1,77 @@
+ignition:
```
ybettan, line 1: Can we somehow reuse `pkg/mcproducer/templates/machine-config.gotmpl`? Do we have to redefine a template?

pkg/mcproducer/mcproducer.go
```diff
@@ -52,29 +46,30 @@ func ProduceMachineConfig(machineConfigName,
...
+	mcfgAPI := mcfg.NewMCFG()
+	jsonIgnition, err := mcfgAPI.GenerateIgnition(kernelModuleImage, kernelModuleName, inTreeModuleToRemove, firmwareFilesPath, workerImageToUse, machineConfigName)
+	if err != nil {
+		return "", fmt.Errorf("failed to create ignition string: %v", err)
 	}
 
-	templateParams["ReplaceInTreeDriverContents"] = base64.StdEncoding.EncodeToString([]byte(scriptReplaceKmod))
-	templateParams["PullKernelModuleContents"] = base64.StdEncoding.EncodeToString([]byte(scriptPullImage))
-	templateParams["WaitForNetworkDispatcherContents"] = base64.StdEncoding.EncodeToString([]byte(scriptWaitForNetworkDispatcher))
+	yamlIgnition, err := yaml.JSONToYAML(jsonIgnition)
```
ybettan, line 55: The mcfgAPI is converting the yaml ignition to a json ignition and here we are re-converting. Can we just return both json and yaml ignitions from the mcfg API? Or don't convert to json to begin with in the mcfg API?

internal/mcfg/mcfg.go
ybettan: What is the difference between MachineConfig and MachineConfiguration?

internal/mcfg/mcfg.go
yevgeny-shnaidman: yes: yaml is more clear and understandable, when looking at the ignition. Unfortunately ignition does not have a well-defined structure, that's why i am using a yaml template

PR #1669, opened 2025-09-15
Fixing status not appear in deprecated PreflightValidationOCP

TomerNewman:
When creating  PreflightValidationOCP with v1beta1 version a new resource with v1beta2 version is created as expected, but the status is only appear in v1beta2 and not in the v1beta1 one.
This is happening because the conversion webhook was not well defined in the CRD.
This commit fixes it by adding into the PreflightValidationOCP CRD a conversion between v1beta1 and v1beta2.

---

/assign @yevgeny-shnaidman @ybettan


ybettan:
/lgtm
@TomerNewman We need to make sure to backport it to the `release-2.5` branch.

PR #1680, opened 2025-10-22
Updating all images to default to the `release-2.5` tag.

ybettan:
We should be able to checkout the release branch and just run `make deploy` or `oc apply -k config/default` and get the release-2.5 version deployed.


ybettan:
/override security
Command injection is in a test file:
```
 ✗ [LOW] Command Injection
   Path: internal/worker/cmdlog_test.go, line 58
   Info: Unsanitized input from a CLI argument flows into os.exec.Command, where it is used as a shell command. This may result in a Command Injection vulnerability.
```

PR #1805, opened 2026-06-03
Release 2.5 - Fix device plugin pod not evicted during node drain

TomerNewman:
When a node is drained, the device plugin DaemonSet pod remains running because DaemonSet pods auto-tolerate NoSchedule taints and the only nodeSelector was the kmm-ready label (which stays while the kmod is loaded). This creates a deadlock: the unloader cannot run while the device plugin holds device files, and the device plugin won't leave because kmm-ready is never removed.

Introduce a new device-plugin-target node label managed by the DevicePluginReconciler. The DaemonSet nodeSelector now requires both kmm-ready AND device-plugin-target. The DevicePluginReconciler watches node taint changes: it adds device-plugin-target to schedulable nodes matching the Module selector, and removes it from unschedulable nodes. This causes the DaemonSet controller to evict the device plugin pod during drain, breaking the deadlock.

---

---

cherrypick [failed](https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1802#issuecomment-4612660544) due to conflicts

### DIFF
--- a/internal/controllers/device_plugin_reconciler.go
+++ b/internal/controllers/device_plugin_reconciler.go
@@ -34,8 +34,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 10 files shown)


TomerNewman:
/override ci/prow/security

failed because of a test file, not related to the KMM code

ybettan:
/hold
We need to wait for https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1803 to merge first. 

Using `/cherry-pick release-2.5` on https://github.com/rh-ecosystem-edge/kernel-module-management/pull/1803 once merged is usually what we do to emphasize that this is an automated cherry-pick. No need to close this one but just for info.

We can merge it once the fix is backported to 2.6

ybettan:
/unhold
It was merged faster than expected :)

PR #1830, opened 2026-06-24
Restrict device plugin hostPath volumes to allowed prefixes

TomerNewman:
Add webhook validation that rejects hostPath volumes in spec.devicePlugin.volumes unless the path resolves under /dev, /sys, /var or /opt. This prevents device plugins from mounting arbitrary host directories such as / or /etc.

---

failed to auto cherrypick due to conflicts


TomerNewman:
/override ci/prow/security
failed on a test file

PR #1831, opened 2026-06-24
Restrict device plugin hostPath volumes to allowed prefixes

TomerNewman:
Add webhook validation that rejects hostPath volumes in spec.devicePlugin.volumes unless the path resolves under /dev, /sys, /var or /opt. This prevents device plugins from mounting arbitrary host directories such as / or /etc.

---

/hold lets wait for #1830
failed to auto-cherrypick due to conflicts

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


TomerNewman:
/override ci/prow/security
failed on a test file

PR #1882, opened 2026-08-23
Retry build/sign resource when its pod was evicted

TomerNewman:
During a cluster upgrade, node drains can delete a Build's pod before it completes, leaving the Build in Error(BuildPodDeleted) forever since Sync() only recreated resources whose spec had changed.
Detect this recoverable failure and delete the resource so it gets retried.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically recreates builds that failed because their build pod was deleted.
  * Reports errors when recovery eligibility cannot be determined or cleanup fails.
  * Treats already-removed resources as successfully cleaned up.
  * Avoids recreating builds with non-recoverable failures.
  * Recovers unchanged resources when their failure state is eligible for restart.

* **Tests**
  * Added coverage for recoverable failures, unsupported resource types, resource changes, cleanup errors, and recovery-check errors.

### DIFF
--- a/internal/buildsign/manager.go
+++ b/internal/buildsign/manager.go
@@ -103,18 +103,19 @@ func (m *manager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage
 		return nil
 	}
 
-	changed, err := m.resourceManager.IsResourceChanged(resource, resourceTemplate)
+	shouldRestart, err := m.resourceManager.ShouldResourceBeRestarted(resource, resourceTemplate)
 	if err != nil {
-		return fmt.Errorf("could not determine if the resource has changed: %v", err)
+		return fmt.Errorf("could not determine if the resource should be restarted: %v", err)
 	}
 
-	if changed {
-		logger.Info("The module's spec has been changed, deleting the current resource so a new one can be created",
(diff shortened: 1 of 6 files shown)


internal/buildsign/manager.go
```diff
@@ -108,13 +108,28 @@ func (m *manager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage
 		return fmt.Errorf("could not determine if the resource has changed: %v", err)
 	}
 
-	if changed {
-		logger.Info("The module's spec has been changed, deleting the current resource so a new one can be created",
-			"name", resource.GetName(), "action", action)
-		err = m.resourceManager.DeleteResource(ctx, resource)
+	deleteReason := "The module's spec has been changed"
```
yevgeny-shnaidman: Suggestion:
lets rename the IsResourceChanged function to ShouldResourceBeRestarted. It will check if the the resouce has changed or is the pod was deleted due to eviction. The Sync function will just call the ShouldResourceBeRestarted, and if it returns true, it will delete the resource. WDYT?

TomerNewman: done

PR #1889, opened 2026-08-27
[release-2.6] Add configurable StartupProbe and LivenessProbe to Module CR

TomerNewman:
Add optional StartupProbe and LivenessProbe fields to CommonContainerSpec, allowing operators to customize probe timing for DevicePlugin containers. When nil, falls back to current defaults.

---


TomerNewman:
/override ci/prow/security
failed on a test file

PR #1899, opened 2026-09-07
Adding network policies for KMM operator

TomerNewman:
Adding Network Policies for KMM and KMM-HUB controllers to be deployed via the makefile and via bundle

---


yevgeny-shnaidman:
we did not had network policies in KMM prior to this commit?

TomerNewman:
These are for the operator not operand

TomerNewman:
/hold
I tested it against 4.21 and 4.12 and it seems to work, lets wait with the merge until we are sure it will pass the release

TomerNewman:
edit, lets wait for feb 2027, releasing until then with NPs is blocked

PR #1912, opened 2026-09-10
Support configurable SecurityContext for DevicePlugin containers

TomerNewman:
Add an optional SecurityContext field to DevicePluginContainerSpec. When set, it overrides the default privileged security context for DevicePlugin containers. When nil, the existing privileged: true default is preserved.

Note: upstream also adds this for DRA containers, but 2.6 does not support DRA, so that part was dropped from this cherry-pick.

---

/assign @yevgeny-shnaidman


TomerNewman:
/override ci/prow/e2e-hub
/override ci/prow/operator-hub-upgrade

fails to install ACM, skipping for now since it blocks the release.

TomerNewman:
/override ci/prow/e2e-hub
/override ci/prow/operator-hub-upgrade

fails to install ACM, skipping for now since it blocks the release.

PR #1913, opened 2026-09-14
Switch to server-side apply for kubectl install/deploy commands

TomerNewman:
Client-side apply stores the full manifest in the
last-applied-configuration annotation, capped at 256KB. The Module and ManagedClusterModule CRDs haven't hit that limit yet, but the pending configurable SecurityContext change (#1323) will push them over it and break kubectl apply.
 Switch to --server-side --force-conflicts preemptively, since it tracks field
ownership on the API server instead and isn't subject to this limit.

---

/cc @yevgeny-shnaidman 
/assign @yevgeny-shnaidman


TomerNewman:
/override ci/prow/e2e-hub

fails to install ACM, skipping for now since it blocks the release.

PR #1918, opened 2026-09-15
Stage modprobedDir in worker pods and set Privileged

NataliShemtov:
When modprobedDir is set, the init container validates and copies
regular files into shared tmp, and both load and unload worker pods
run Privileged. When modulesLoadingOrder is set, softdep.conf is
always a file-level mount so /etc/modprobe.d stays writable.

This is a manual cherry pick for this [Upstream](https://github.com/kubernetes-sigs/kernel-module-management/pull/1345)

Upstream-Commit: 036ff6b24884bcc9c83f6d0d6b38db9a22484351


NataliShemtov:
opened a manual one since there was an issue with #1917

PR #1934, opened 2026-09-22
Introducing InitramfsModule CRD

yevgeny-shnaidman:
This CRD will be used in the initramfs upgrade scenario


api/v1beta1/initramfsmodule_types.go
```diff
@@ -0,0 +1,118 @@
...
+
+// InitramfsNodeStatus is the observed initramfs state for one node.
+type InitramfsNodeStatus struct {
+	Name string `json:"name"`
+
+	// +optional
+	KernelVersion string `json:"kernelVersion,omitempty"`
+
+	InitramfsCreated bool `json:"initramfsCreated"`
+	InitramfsApplied bool `json:"initramfsApplied"`
```
TomerNewman: what about InitramfsFailed?

TomerNewman: maybe instead of bools you can use status.conditions?

api/v1beta1/initramfsmodule_types.go
```diff
@@ -0,0 +1,118 @@
...
+// InitramfsModuleSpec describes out-of-tree kernel modules to place into initramfs.
+type InitramfsModuleSpec struct {
+	// Selector limits which nodes receive the generated initramfs.
+	// Empty or omitted targets worker nodes.
+	// +optional
+	Selector map[string]string `json:"selector,omitempty"`
+
+	// ModuleNames are the OOT kernel modules to extract and load
+	// (.ko basename without .ko).
+	// +kubebuilder:validation:MinItems=1
```
TomerNewman, line 33: I would add 
// +kubebuilder:validation:UniqueItems=true
// +kubebuilder:validation:items:Pattern=`^[A-Za-z0-9_-]+$`

TomerNewman (approved):

